PR #74, opened 2015-06-09
Update users_sf2gh.json

Pauldenny:
Not sure if this is going to add my username or not


cmungall:
Oops, look like this introduced a conflict. Sorry about that, I added you separately @Pauldenny

PR #78, opened 2015-06-10
Update users_sf2gh.json

asangrador:


cmungall:
I added @asangrador outside this PR

PR #89, opened 2015-07-29
Update users_sf2gh.json

cooperl09:
Added GitHub ID for Laura Moore


kltm:
@cmungall can we go ahead and close this one? I'm assuming it's in there already now that the transition is done.

cmungall:
I'll merge, may be useful for getting these into the users file, there's overlap with the PO folks

PR #122, opened 2015-11-10
Update users.yaml

tberardini:
Added ORCID for Donghui.

### DIFF
--- a/metadata/users.yaml
+++ b/metadata/users.yaml
@@ -1179,8 +1179,9 @@
       allow-write: true
       allow-freeform: true
   xref: 'GOC:dhl'
-  uri: 'GOC:dhl'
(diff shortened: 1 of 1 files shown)


cmungall:
Not sure we need the comment field but does no harm

kltm:
Probably copying the legacy of the comment field containing the orcid--there's a lot of that still in there.

tberardini:
Yup, I just copied what was in my record over.

kltm:
No harm, but like @cmungall sez, not necessary anymore. In a previous version of this data, before it all got merged together, the orcid was kept in the comment field when available. After all the different sources were merged into this new single format, since a lot of people seemed to be using the comment field for different things, we just left it alone. This leads to a little replication, but not a big deal.

PR #130, opened 2016-01-04
Update users.yaml

paolaroncaglia:


metadata/users.yaml
```diff
@@ -157,6 +157,11 @@
       allow-write: true
       allow-freeform: true
 -
+uri: 'http://orcid.org/0000-0002-3898-1727'
+  xref: 'GOC:bc’
```
kltm: @paolaroncaglia It looks like there is a backtick here instead of a quote, which would be bad, and the reason I'm going to reject this PR. I'll try and fix the tests to catch this, and could you resubmit with the single quote instead of the tick?

kltm:
@paolaroncaglia It looks like there is a backtick here (line 161) instead of a quote, which would be bad, and the reason I'm going to reject this PR. I'll try and fix the tests to catch this, and could you resubmit with the single quote instead of the tick?

PR #139, opened 2016-01-08
Fixed ISBN link, and added Planteome_gene generic reference stanza

austinmeier:
Thanks!

### DIFF
--- a/metadata/db-xrefs.yaml
+++ b/metadata/db-xrefs.yaml
@@ -983,7 +983,7 @@
   entity_types:
     - type_name: entity
       type_id: BET:0000000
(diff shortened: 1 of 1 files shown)


kltm:
@cmungall this looks technically fine, but before merging I was wondering if you had any opinion about the google search from line 1976?

cmungall:
Merging, but this is a bit of a red flag, @austinmeier the example is from maizegdb, how would this become orphaned?

austinmeier:
MaizeGDB was a bad example.  That was my fault, I just had that one handy in the editor.  I will speak with Pankaj, and see which generic references need links.

PR #151, opened 2016-01-22
Update users.yaml

paolaroncaglia:


paolaroncaglia:
Adding dbxref for Giulia Antonazzo (Flybase)

PR #158, opened 2016-02-05
added TO_GIT, PO_GIT, EO_GIT, and PSO_git dbxref stanzas

austinmeier:
We will eventually transition to using the URL for these issue trackers, but would like them to work in the current state.

### DIFF
--- a/metadata/db-xrefs.yaml
+++ b/metadata/db-xrefs.yaml
@@ -899,13 +899,6 @@
   entity_types:
     - type_name: entity
       type_id: BET:0000000
-- database: GermOnline
-  name: GermOnline
-  generic_urls:
-    - http://www.germonline.org/
-  entity_types:
-    - type_name: entity
-      type_id: BET:0000000
 - database: GO
   name: Gene Ontology Database
   generic_urls:
--- a/metadata/db-xrefs.yaml
+++ b/metadata/db-xrefs.yaml
@@ -759,7 +759,7 @@
 - database: EO_GIT
   name: GitHub
   generic_urls:
-    -https://github.com/Planteome/plant-environment-ontology
+    - https://github.com/Planteome/plant-environment-ontology
   entity_types:
     - type_name: entity
       type_id: BET:0000000
@@ -2054,7 +2054,7 @@
 - database: PO_GIT
   name: GitHub
   generic_urls:
-    -https://github.com/Planteome/plant-ontology
+    - https://github.com/Planteome/plant-ontology
   entity_types:
     - type_name: entity
       type_id: BET:0000000
@@ -2180,7 +2180,7 @@
 - database: PSO_GIT
   name: GitHub
   generic_urls:
-    -https://github.com/Planteome/plant-stress-ontology
+    - https://github.com/Planteome/plant-stress-ontology
   entity_types:
     - type_name: entity
       type_id: BET:0000000
@@ -2709,7 +2709,7 @@
 - database: TO_GIT
   name: GitHub
   generic_urls:
-    -https://github.com/Planteome/plant-trait-ontology
+    - https://github.com/Planteome/plant-trait-ontology
   entity_types:
     - type_name: entity
       type_id: BET:0000000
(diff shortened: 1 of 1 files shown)


cmungall:
I recommend moving away from this way of doing it. Just use either full URLs, or have a generic

github:ORG/REPO/issues/NUMBER

ID/shorthand

cooperl09:
Posted a test of using the url, but it is not linking back

See example on the mappings tab: http://dev.planteome.org/amigo/term/TO:0000283

kltm:
@cooperl09 Did you have the install scripts re-reun on that server (@elserj)?

kltm:
@cooperl09 Also, if you're using the the go-site master version, until it is merged it would not show up in a standard re-install; one would have to force the installer to use the local copy.

elserj:
Install scripts were rerun, but are still using the master go-site dbxref file.  Looks like the PR request failed, so that will need to be fixed first.

However, putting in the URL directly doesn't seem to work, at least the way I would expect it to.  I believe the problem is that the parser is interpreting "http:" (or "https") as a dbxref and trying to use the rest of the URL as an argument.  Please check Lol's example http://dev.planteome.org/amigo/term/TO:0000283
in the mappings.  

Is there a different way the URL should be put in the OBO file?  We have it in as:
xref: https://github.com/Planteome/plant-trait-ontology/issues/378

kltm:
@elserj I'm not sure what the desired outcome is for your changes. Could you further explain what you difference you'd expect to see with your change?

kltm:
@elserj Whoops--saw it, let me take a look.

elserj:
I would expect it to be an actual clickable link in the "mapping" section.  It looks like it is trying to break it up because the URL has a (required) ":" in it.  It then treats http as the dbxref, like TAIR, and the rest of the URL as the argument, as that is the format for all other xrefs.

If that makes it clearer.

cooperl09:
@kltm Any idea why the Travis check failed on @austinmeier 's PR?  We would prefer to use those tags as the DBXrefs sincew they are all in the files already.

austinmeier:
I figured out what happened:  There needed to be a space between the "-" and the URL.  
Also, there seemed to be a stanza for 'GermOnline' that was duplicated... No idea how that happened, as I did not add or modify that stanza.  It looks like the checks have passed now.

kltm:
Does this work better now?

cooperl09:
@kltm Not sure what you are referring to above.... url is not working still

kltm:
Will go ahead with the merge.

elserj:
Just did a "gulp install" on dev, TO_GIT and I assume the others are working now, but the http(s) still doesn't.  I know @cmungall suggested that was the preferred method going forward, but it doesn't seem to work for now, unless we are doing something incorrect.

kltm:
@elserj 
Okay, I've done more digging; the issue is that we are running into this:
```
"database_xref":["http://github.com/Planteome/plant-trait-ontology/issues/378"],
```
Instead of the (likely) expected:
```
"database_xref":["TO_GIT:378"],
```
could you check that you're loading the right thing at your end.

elserj:
Sorry, I think things got a bit confused here as there are actually two issues being discussed in this tracker. 

The original issue, where the TO_GIT, EO_GIT, etc were added to the dbxrefs.yaml file has been completed as of your merge commit.

However, in the first reply comment (second comment), @cmungall suggested that rather than use TO_GIT as the dbxref, we move to a full URL instead.  This suggestion was also made it in another issue ticket somewhere, but I don't remember which one.

The rest of the issue is then us testing out using a full URL.

So, these refs now work:
xref: TO_GIT:378
as the original pull request was attempting to fix.

However, we also started testing out whether a full URL would work as suggested by Chris, but these do not appear to be working.  We only have one test xref in the following format:
xref: https://github.com/planteome/plant-trait-ontology/issues/378

Now, we are testing this in an OBO file, so maybe it works as expected in an OWL, but I haven't tested that yet.  We were mostly just trying to point out that the suggestion of using full URLs doesn't work, probably because of the colon after the http(s) being interpreted as a separator in the dbxref, unless of course we are doing something wrong.

If you want, we can set up a call sometime today to actually discuss this if it still isn't clear.

kltm:
I'll leave that to @cmungall - the way the db-xref codes works is fully consistent with the expectation that IDs are of the for namespace:id. I'm not sure what should be done here if the pattern is disrupted (as this is a fairly base assumption) in AmiGO.

PR #167, opened 2016-03-09
Added IRIC stanza for International Rice Informatics Consortium

austinmeier:
International Rice Research Institute germplasm data is being added.  Would like to link back to oryzasnp


kltm:
@austinmeier The given example URL does not resolve and the host oryzasnp.org does not seem to be pingable.

kltm:
@austinmeier I take it back, it resolved after a tremendous wait.

austinmeier:
Yeah, I'll speak with the developer here, and see if there is a reason for the slow load.  Thanks for making the merge.

PR #199, opened 2016-07-26
Adding Istvan

cmungall:
This is for the OBO Noctua instance

### DIFF
--- a/metadata/users.yaml
+++ b/metadata/users.yaml
@@ -2354,7 +2354,7 @@
   email-md5:
     - 7a113f413bef0385f3f26ea11500697e
   authorizations:
-    noctua-obo:
+    noctua-go:
       allow-edit: true
   organization: PhenotypeRCN
 -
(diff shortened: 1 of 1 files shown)


kltm:
There is no concept of "noctua-obo". If you think that users.yaml should contain users with incompatible profiles, we'll have to make some changes to how barista functions.

The likely format would have to change to something more like:
```yaml
nickname: "Istvan Miko"
uri: 'http://orcid.org/0000-0001-9719-0215'
email-md5:
   - 7a113f413bef0385f3f26ea11500697e
authorizations:
   noctua:
      allow-namespaces: ['obo']
      allow-edit: true
organization: PhenotypeRCN
```

And each barista would have to know which which namespaces it controlled. There is a lot of wiring in there as well for barista to handle multiple backends, so depending on what hat you were wearing, you could send commands to different baristas. However, that's probably more than intended or desired here.

PR #211, opened 2016-08-16
Updated the stanza for MaizeGDB references

cooperl09:
Still had the old syntax


kltm:
@cooperl09 It looks like you spacing is incorrect in the YAML, causing an error.
YAML is like python, where space is significant.

cooperl09:
See if this works better

cooperl09:
Thanks Seth!

Issue #214, opened 2016-08-24
Move GO_REFs over to go-site


cmungall:
Currently exist in svn, rendered directly as http://www.geneontology.org/doc/GO.references

And awful: http://www.geneontology.org/cgi-bin/references.cgi

move each ref to its own md/yaml file, analogous to obofoundry. Use default github rendering, optional jekyll liquid templates later

cc @kltm @monicacecilia

kltm:
What's the endgame here? Are these individual blobs going to be pushed somewhere to resolve outside of GH? Are we going to have something like http://references.geneontology.org/GO_REF:0000001 ?

cmungall:
Minimally, db-xrefs.yaml can be changed to point at the github URLs. A bit ugly but no worse than currently.

There are a variety of rendering options beyond this:

 * Static site with GitHub pages + jekyll, a la OBO
 * Static site generated using jekyll but deployed independent of GH (e.g. direct copy to S3)
 * Custom rendering in amigo (could use [Text::Markdown](http://search.cpan.org/~bobtfish/Text-Markdown-1.000031/lib/Text/Markdown.pm) but not really any need for this yet as we don't have much markdown in the abstract
 * Quick pandoc pipeline run in jenkins (but quickly run into questions about what css to use etc)

cmungall:
As for the purl, I think an obolibrary purl, e.g. $OBO/go/references/0000001

cmungall:
The metadata is now sourced from github

Still to change existing site links

mcourtot:
@tonysawfordebi mentioned that http://purl.obolibrary.org/obo/go/references/0000001 doesn't resolve. Should we just create a new entry under https://github.com/OBOFoundry/purl.obolibrary.org/tree/master/config? I'm a bit wary of mixing this with the OBO prefixes. @jamesaoverton?

jamesaoverton:
Any PURL starting with `http://purl.obolibrary.org/obo/go/` can be defined in <https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/go.yml>. I think <http://purl.obolibrary.org/obo/go/references/0000001> is a perfectly good PURL to define.

tonysawfordebi:
When the PURLs are available, the GO_REF entry in https://github.com/geneontology/go-site/blob/master/metadata/db-xrefs.yaml should also be updated.

cmungall:
yes, will need changed in 2 places

mcourtot:
[...]

PR #215, opened 2016-08-24
Seeding goref YAML/MDs

cmungall:
See #214


cmungall:
This is how it looks: https://github.com/geneontology/go-site/blob/issue-214/metadata/gorefs/goref-0000002.md

A bit ugly, but we could easily pandocify in advance of jekyll or render any number of ways

cmungall:
@kltm ok to merge?

kltm:
Checks failed?

kltm:
@cmungall still failing.

cmungall:
time to switch to https://pypi.python.org/pypi/pykwalify

kltm:
You'll have to run that through virtualenv.

kltm:
@cmungall The failure
```
The command "./scripts/combine-datasets-metadata.py metadata/datasets/*.yaml > metadata/datasets.json" exited with 127.
```
would seem to be unrelated?

cmungall:
Not following. The error is due to a lack of python3 in the ruby build environment. Would virtualenv make it easier to combine the? Seems more straightforward to just go for a straight python environment, it's what we use for most travis jobs.

kltm:
IIRC, the "build environment" in Travis is just the base image and some var/env stuff over a pretty vanilla Ubuntu image. You can install any package you want and run any command within that environment, including python3. (I think that's right, last time I checked something like this in detail was in the GitLab CI docs, so a shout-out there, but a quick look at the Travis docs seems to agree.)
Of course, python3 is not standard in base Ubuntu, so it must be installed separately and referred to explicitly over the default python2. As well, pykwalify is just (apparently) a random library and not packaged in any standard location, so must be installed separately either with pip directly or using virtualenv locally--a bit of an irritation. 
As far as the current job goes, it does not necessarily need to be the ruby image, that's just the way I started with it--there are other ways to skin that cat.
The reason I default to kwalify is not any deep love, but that it is a standard package in Ubuntu, making upgrading, security, etc. easy for our systems; the Travis CI is just an extension of that standardization, but can be changed as their VMs are throw-away.

Issue #220, opened 2016-08-31
EnsemblPlants xref in db-xrefs.yaml doesn't work correctly


elserj:
After realizing we had it as Ensembl_Plants instead of EnsemblPlants as in the db-xrefs.yaml file, we switched it in the gaf file and reloaded.  However, now the URL that is generated is:
http://www.ensemblgenomes.org/id/[example_ID]
when it should be:
http://www.ensemblgenomes.org/id/EMT11640 (for example)

My guess is that all the Ensembl* entries in the file are [example_ID] rather than [example_id].  I'm guessing the case matters, and so AmiGO is not replacing [example_ID] with what it should be.

cmungall:
Looks like @austinmeier fixed this one

PR #221, opened 2016-08-31
changed case of [example_ID] to [example_id] for ensemblPlant stanza

austinmeier:
Links generated using this stanza were directing to the URL:  "http://ensemblgenomes.org/id/[example_ID]"  instead of properly substituting the ID in the URL.  I don't know if the issue is related to the case of the "ID" in the url syntax, but we would like to fix this if it is.

See also:  https://github.com/geneontology/go-site/issues/220

### DIFF
--- a/metadata/db-xrefs.yaml
+++ b/metadata/db-xrefs.yaml
@@ -746,7 +746,7 @@
   entity_types:
     - type_name: gene
       type_id: SO:0000704
-      url_syntax: http://www.ensemblgenomes.org/id/[example_ID]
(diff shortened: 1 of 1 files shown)


elserj:
I believe the issue affects more than just EnsemblPlants, might as well also fix EnsemblFungi, EnsemblMetazoa, and EnsemblProtists.

kltm:
Are we waiting for more fixes, or should we take the PR?

austinmeier:
All fixes have been made.  Merge away!

elserj:
@kltm, we had to deal with an issue where our fork was out of date and missing the others. We should be good to go now.

Issue #237, opened 2016-10-21
Switch from ruby kwalify to something better supported


cmungall:
Continued from #214 

Kwalify is strange and not well supported. It would also be better to switch the travis to being python as default. The python env mixes well with java in travis, and this covers most of our tooling.

For a replacement for kwalify, possibly json schema or shacl (itself not well supported but fits into our future plans well)

cmungall:
Note: marshmallow to json-schema is supported: https://github.com/fuhrysteve/marshmallow-jsonschema

kltm:
That seems pretty python-specific.

cmungall:
Note we're also using JSON schema to validate the (yaml) DOSDPs in go-ontology

PR #238, opened 2016-10-21
DO NOT MERGE YET Python port

cmungall:
See #237

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,9 @@
-language: ruby
-rvm: "1.9.3"
+language: python
+python:
+  - "3.4"
 
-# before_install:
-#   - sudo apt-get update -qq
-#   - sudo apt-get install -qq kwalify
-## Possible other way.
 install:
-  - gem install kwalify
+  - pip3 install pykwalify
 script:
   ## Quick user data check.
   - kwalify -E -m metadata/users.schema.yaml 2>&1 | tee out.log && grep 'INVALID\|ERROR' out.log; test $? -ne 0
@@ -17,6 +14,7 @@ script:
   - kwalify -E -f metadata/db-xrefs.schema.yaml metadata/db-xrefs.yaml 2>&1 | tee out.log && grep 'INVALID\|ERROR' out.log; test $? -ne 0
   - kwalify -E -m metadata/datasets.schema.yaml 2>&1 | tee out.log && grep 'INVALID\|ERROR' out.log; test $? -ne 0
   - kwalify -E -f metadata/datasets.schema.yaml metadata/datasets/*.yaml 2>&1 | tee out.log && grep 'INVALID\|ERROR' out.log; test $? -ne 0
+  - ./scripts/combine-datasets-metadata.py metadata/datasets/*.yaml > metadata/datasets.json
 notifications:
   email:
     - sjcarbon@lbl.gov
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,14 +6,7 @@ install:
   - pip3 install pykwalify
 script:
   ## Quick user data check.
-  - kwalify -E -m metadata/users.schema.yaml 2>&1 | tee out.log && grep 'INVALID\|ERROR' out.log; test $? -ne 0
-  - kwalify -E -f metadata/users.schema.yaml metadata/users.yaml 2>&1 | tee out.log && grep 'INVALID\|ERROR' out.log; test $? -ne 0
-  - kwalify -E -m metadata/groups.schema.yaml 2>&1 | tee out.log && grep 'INVALID\|ERROR' out.log; test $? -ne 0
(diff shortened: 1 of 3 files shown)


cmungall:
OK, it seems pykwalify operates differently

```
pykwalify --data-file metadata/groups.yaml --schema-file metadata/groups.schema.yaml
 ERROR - validation.invalid
 ERROR -  --- All found errors ---
 ERROR - ["Value 'http://lbl.gov' does not match pattern '/^[\\/\\.\\-\\:\\w]+$/'. Path: '/0/id'", "Value 'http://geneontology.org' does not match pattern '/^[\\/\\.\\-\\:\\w]+$/'. Path: '/1/id'", "Value 'http://planteome.org' does not match pattern '/^[\\/\\.\\-\\:\\w]+$/'. Path: '/2/id'", "Value 'http://berkeleybop.org' does not match pattern '/^[\\/\\.\\-\\:\\w]+$/'. Path: '/3/id'"]
T```

**Comment by cmungall** [comment]
OK, so it appears pykwalify does not like the `//` surrounding a regex

pykwalify gives super-informative error messages like:

```
Traceback (most recent call last):
  File "/usr/local/bin/pykwalify", line 9, in <module>
    load_entry_point('pykwalify==1.5.1', 'console_scripts', 'pykwalify')()
  File "/usr/local/lib/python3.4/site-packages/pykwalify/cli.py", line 84, in cli_entrypoint
    run(parse_cli())
  File "/usr/local/lib/python3.4/site-packages/pykwalify/cli.py", line 71, in run
    c.validate()
  File "/usr/local/lib/python3.4/site-packages/pykwalify/core.py", line 147, in validate
    self._start_validate(self.source)
  File "/usr/local/lib/python3.4/site-packages/pykwalify/core.py", line 192, in _start_validate
    self._validate(value, root_rule, path, done)
  File "/usr/local/lib/python3.4/site-packages/pykwalify/core.py", line 209, in _validate
    self._validate_sequence(value, rule, path, done=None)
  File "/usr/local/lib/python3.4/site-packages/pykwalify/core.py", line 337, in _validate_sequence
    val = V[v]
KeyError: 'xref'
```

**Comment by kltm** [comment]
Can you pin this to a version so an upstream change does not affect the unit tests?

**Comment by cmungall** [comment]
ok figured the issue. Looks like pykwalify is ignoring `required: false` if `unique: true` is set.

I don't know where the canonical spec for kwalify is, but the pykwalify interpretation is not unreasonable. However, it differs from the kwalify one. Sigh

May be related:

https://github.com/Grokzen/pykwalify/issues/66

**Comment by Grokzen** [comment]
@cmungall Hi. I saw this issue was linked in my repo and i thought i would shed some light on the issues you posted in here.

The main error i see in this thread is about the regex validation and that they are not compatible between ruby and python. I do not really know why ruby implements regex the way they do but you can at least read up on it here if you want https://ruby-doc.org/core-2.4.0/Regexp.html. The original kwalify source code executes this line when it does pattern validation https://github.com/sunaku/kwalify/blob/master/lib/kwalify/validator.rb#L224

The solution to this regex problem is easy. I have seen this regex problem before and from what i have seen when people use pykwalify and they are not porting an existing schema from ruby to python is that there is no issue because wrapping a python style regex with slashes makes no sense.

I implemented a new flag to pykwalify that can help to make the transition a bit easier if there is a problem with changing the schema. The commit is here https://github.com/Grokzen/pykwalify/commit/f93a4ca4581cb439187c0dc38f925fa6a9760d8c and with this i can run the following based on the files and content i found in your repo

```
(pykwalify3.4) in ~/code/pykwalify on unstable)
$ pykwalify -d groups.yaml -s groups.schema.yaml --fix-ruby-style-regex
 INFO - validation.valid
```

where if it would be without the flag i get

```
(pykwalify3.4) in ~/code/pykwalify on unstable)
$ pykwalify -d groups.yaml -s groups.schema.yaml
 ERROR - validation.invalid
...
pykwalify.errors.SchemaError: <SchemaError: error code 2: Schema validation failed:
 - Value 'http://lbl.gov' does not match pattern '/^[\/\.\-\:\w]+$/'. Path: '/0/id'.
 - Value 'http://geneontology.org' does not match pattern '/^[\/\.\-\:\w]+$/'. Path: '/1/id'.
 - Value 'http://planteome.org' does not match pattern '/^[\/\.\-\:\w]+$/'. Path: '/2/id'.
 - Value 'http://berkeleybop.org' does not match pattern '/^[\/\.\-\:\w]+$/'. Path: '/3/id'.
 - Value 'http://zfin.org' does not match pattern '/^[\/\.\-\:\w]+$/'. Path: '/4/id'.: Path: '/'>
```

This will be in the next major release of the tool.

On the other error you found
```
KeyError: 'xref'
```
i would love to know what files/data you used to get that error so i can take a stab on that one also because that error is not something that i see that often.

On a side note, the problem is not realted to PR 66 in pykwalify repo becuase the problem in that issue is regarding partial schema:s and include keyword and not the required/unique keywords per say.

PR #242, opened 2016-10-27
Update query-github-for-new-tickets.js

mcourtot:
Removed extra type


mcourtot:
ok - after testing type is created or updated - confusing with the issue type, but this PR should not be merged.

PR #245, opened 2016-11-01
PO custom references link

cooperl09:
Updated URL syntax for links to the PO custom references which have been moved to the Planteome site.


cooperl09:
Thanks Seth!

cmungall:
Just FYI, I see you're using drupal to manage the refs, we're doing this directly in github for GO, see https://github.com/OBOFoundry/purl.obolibrary.org/pull/258

PR #252, opened 2016-11-06
Added zfin to groups.yaml

doughowe:
Simple addition of ZFIN to groups.yaml


kltm:
@doughowe Now you need to add the groups section to your users.yaml :)

PR #273, opened 2016-11-29
made github account a first-class parameter

nathandunn:
Not sure if you still need accounts @kltm , but this is the right change to fix https://github.com/geneontology/termgenie/issues/98 with other fixes.

### DIFF
--- a/metadata/users.yaml
+++ b/metadata/users.yaml
@@ -53,8 +53,8 @@
       allow-freeform: true
       allow-management: false
   uri: 'http://orcid.org/0000-0002-4862-3181'
+  github: nathandunn
   accounts:
-    github: nathandunn
     orcid: '0000-0002-4862-3181'
     email-md5:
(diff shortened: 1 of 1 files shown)


cmungall:
I don't understand what you're doing here, is this an agreed upon change to the schema? We need to decide whether github and other accounts are top level or not. But it shouldn't be in both places.

nathandunn:
Definitely doesn't need to be in both places but I don't know the history of the schema.  Was just trying to quickly fix the termgenie code.  However, if you think that this will violate it, I can definitely work around it.  Just let me know.  

Nathan

cmungall:
well it's already violating the schema, hence the travis err

I think the code will have to be fixed to do the lookup in the accounts object

nathandunn:
No problem.  Please close it then. 

Nathan

nathandunn:
Talked to @cmungall  and decided not to do it this way.

PR #276, opened 2016-12-01
Update users.yaml

tberardini:
Added GH id and ORCID for Leonore Reiser, also other little edits.

### DIFF
--- a/metadata/users.yaml
+++ b/metadata/users.yaml
@@ -1326,7 +1326,7 @@
   uri: 'http://orcid.org/0000-0003-3335-4537'
   organization: TAIR
   accounts:
-    github: 'dhldhl'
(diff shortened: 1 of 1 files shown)


kltm:
@tberardini After making a PR, it's good to stick around and see if it passes the tests.
In this case, the schema tests fail here on the uri for Lenore being a dupe; best to merge those records and make another PR.

tberardini:
Whoops.  I don't know what to do next. Start all over again?  How do I edit the version that I started working on? Can you send link?  Thanks.

kltm:
There seems to be only 1 "real" change, but you'll need to merge with the other Lenore info anyways, so I think I would probably just close this one and start over.

Issue #282, opened 2016-12-06
Convert old annotation_qc.xml


cmungall:
A while ago we agreed on xml for specifying our rules
http://www.geneontology.org/quality_control/annotation_checks/annotation_qc.xml

These are translated to HTML placed on drupal, and a hyperlink provided for each rule:
http://geneontology.org/page/annotation-quality-control#GO_AR:0000014

These URLs are provided in the owltools GAF checks

However, these have suffered some bitrot:

 - the HTML generation does not work well with drupal
 - the EBI ruleset has diverged cc @tonysawfordebi See also #123 
 - some rule implementations are SQL
 - we never really added to these
 - we want to use the same system for noctua/lego rules

We should switch to the same system as https://github.com/geneontology/go-site/tree/master/metadata/gorefs

We can have a yaml structured blob inside human-readable markdown.

TODO

 - [ ] decide on rough format
 - [ ] convert xml to md or json (preserve IDs?)
 - [ ] ensure EBI set covered
 - [ ] rewrite owltools GAF reports
 - [ ] make some rules executable (will discuss this with @balhoff - consider using SPARQL or SHACL)

cmungall:
Currently on a branch, can be viewed here for now:

https://github.com/geneontology/go-site/tree/qc-rules/metadata/rules

kltm:
I think it's (more than) a little weird to have a YAML blob in the markdown--this is just the creation of another format. It would be also just much that harder to parse and control.

I think just straight YAML, with a large text/markdown blob as the final field would be more appropriate. YAML very much supports extended and easy-to-manage text fields.

kltm:
E.g. `>` and `|` field controls.

cmungall:
See also #287

cmungall:
We need ASAP a quick view over all rules. I can do a first pass at the script.

cmungall:
Some of the reasons for doing it this way:

  - tried and tested with the OBO Foundry system (works well with 
jekyll)
  - this is a documentation format first, with metadata embedded
  - default rendering works well for humans
  - extraction is trivial for computational purposes

kltm:
You write it and it's here, so that counts for a lot. However:

- is there a validator for this new format?
- the jekyll connection is really about tying it to github at some level--I don't think you'd want to start bringing on ruby dependencies otherwise
[...]

PR #283, opened 2016-12-06
Annotation Qc rules

cmungall:
First pass at #282


cmungall:
until this is merged, browser here: https://github.com/geneontology/go-site/tree/qc-rules/metadata/rules

e.g

https://github.com/geneontology/go-site/blob/qc-rules/metadata/rules/gorule-0000002.md

PR #285, opened 2016-12-08
New users

cmungall:
https://github.com/geneontology/go-ontology/issues/12844


metadata/users.yaml
```diff
@@ -225,8 +225,7 @@
   nickname: 'Jessica Buxton'
   organization: BHF-UCL
   comment: 'status inactive, Collaborator with the BHF group at UCL'
- 
--
+ -
```
kltm: @cmungall No space here

kltm:
@cmungall Shall I close this then? No longer needed?

PR #296, opened 2016-12-27
Added GRIMS stanza for IRRI's GRIMS data to link to Genesys

austinmeier:


kltm:
@austinmeier 
This looks clean to me. A quick question though--the path you gave seems to just be a generic search, rather than direct information about the entity (e.g. something like https://www.genesys-pgr.org/acn/id/607435 ); is this correct?

austinmeier:
Yes.  Unfortunately, the IDs that we have are the PI number of the accession, not the actual ID that Genesys has assigned.  Their URLs use their ID, not the PI number, so sending them to the search is the best I can do.  Just one more click for users.

kltm:
I'll merge right now, but I want to keep @cmungall in the loop.

austinmeier:
Sounds good to me.  Thanks

PR #297, opened 2017-01-05
Added stanza for GRIN descriptors

austinmeier:
I'm not sure if there is a more elegant way to add this as a subsection of the GRIN database stanza, so I just added it in the only way I know how.  Let me know if there is a protocol for this sort of thing.
Thanks


kltm:
`synonyms` should be additional database ids; what you have looks like a description, so I'd change that or leave it out.

This looks like a pretty standard entry to me--is there something that I should be looking for?

cmungall:
+1

ideally GRIN would provide a universal resolver for descriptors and germplasm, and would have local IDs discriminating these. But I don't think they do, so this is fine

austinmeier:
Okay, you're right.  I removed the synonyms from the new stanza to avoid confusion.

PR #317, opened 2017-03-08
Switch html to md in rules

### LINKED ISSUE #282: Convert old annotation_qc.xml

cmungall:
Fixes #282

### DIFF
--- a/metadata/rules/README.md
+++ b/metadata/rules/README.md
@@ -29,17 +29,24 @@
  * status: Implemented
 
 
+The following basic checks ensure that submitted gene association files
+conform to the GAF spec, and come from the original GAF check script.
 
-<p>The following basic checks ensure that submitted gene association files conform to the GAF spec, and come from the original GAF check script.</p>
-<ul><li>Each line of the GAF file is checked for the correct number of columns, the cardinality of the columns, leading or trailing whitespace</li>
-<li>Col 1 and all DB abbreviations must be in <a href="http://www.geneontology.org/cgi-bin/xrefs.cgi">GO.xrf_abbs</a> (case may be incorrect)</li>
-<li>All GO IDs must be extant in current ontology</li>
-<li>Qualifier, evidence, aspect and DB object columns must be within the list of allowed values</li>
-<li>DB:Reference, Taxon and GO ID columns are checked for minimal form</li>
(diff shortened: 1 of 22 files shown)


cmungall:
For the record, here is how I did this. A one-off so not checking in:

```
#!/usr/bin/perl -w
use strict;

foreach my $f (@ARGV) {
    open(F,$f) || die $f;
    my $yaml = 0;
    my $block = "";
    my $out = "";
    while(<F>) {
        if (m@^\-\-\-@) {
            if ($yaml) {
                $yaml = 0;
            }
            else {
                $yaml = 1;
                $block = "";
            }
        }
        else {
            if ($yaml) {
                $block .= $_;
            }
            else {
                $out .= "$_";
            }
        }
    }
    if ($yaml) {
        die "unbalanced yaml markers";
    }
    close(F);
    open(F, ">$f.tmp");
    print F "---\n$block---\n";
    open(Z,">tmp.html");
    print Z $out;
    close(Z);
    print `pandoc tmp.html -o tmp.md`;
    open(Z,"tmp.md") || die;
    while(<Z>) {
        print F $_;
    }
    close(F);
    print `mv $f.tmp $f`;
    close(Z);
    
}

exit 0;

```

PR #318, opened 2017-03-13
Added summary README for gorefs. 

### LINKED ISSUE #271: Provide view over all gorefs
See https://github.com/geneontology/go-site/issues/214#issuecomment-241926666

Note for the typical end user a per-ref page is sufficient

For curators it is useful to see everything on one page as before, to avoid adding dupes.

cmungall:
Fixes #271


hattrill:
Hi Chris, Could you add the external accessions to this page as well?

PR #326, opened 2017-03-20
if the yaml config has 'aggregates', then we pass it through as a megafile, doing limited operations

dougli1sqrd:


dougli1sqrd:
Oh jeez! Let me fix that actually! Let me rename your `all()` function to something else to not collide with the builtin `all()`.

PR #330, opened 2017-04-10
ansible now provisions and maintains blazegraph sparql endpoint

dougli1sqrd:
This is ansible that will run and update the blazegraph sparql endpoint. If a new journal file is available when run, it'll download the new one, replace the old one, and restart blazegraph. We'll need to discuss an actual deploy strategy for getting this on a real server and how that works. We can make this run now if we want, but as we go along I'd like to continue to containerize this.


kltm:
The new ansible stuff is being built up in geneontology/operations. There are still open questions about what parts to use docker on and where to avoid it. We are also looking at switching over to systemd for job control (although may remain with supervisord if we can get it to behave properly in production).

I have never managed to get Vagrant to work reliably on my local machine, so am uncomfortable about having it as the set test rig, but this is rather minimal and likely not an issue. We can talk more about that. (precise is a rather old distro as well)

dougli1sqrd:
@cmungall I'll write up a readme. 

@kltm We should feel free to change whatever image is being used for Vagrant. You're right, that was just for testing, so it's very flexible. In the Vagrant/Ansible tutorial, that was just what they were using.

kltm:
I won't be changing any Vagrant files as I can't run them locally to test. We should be standardizing everything onto 16.04 for now. I know practically that the libc version causes trouble at least between 14.04 and 16.04, so there may be other gotchyas floating around.

cmungall:
is this still relevant?

kltm:
@cmungall @dougli1sqrd I do not know.

cmungall:
what is our plan for deploying blazegraph? how does it relate to how we do the equivalent one in monarch cc @yy20716

kltm:
I believe this is no longer relevant. Deployments (real or docker) will be copying onto machine filesystem from skyhook and restarting. The exact formulation (docker tmpfs, volume, rebuild, pull in, etc.) has not been worked out and will depend on whatever is easiest/most robust to puppet from the pipeline.

PR #332, opened 2017-04-10
adding PLANA_REF

srobb1:
I would like to add our planarian reference pages for ontology terms that have no universal IDs to this yaml.

I followed the directory structure for [GO_REF go-site](https://github.com/geneontology/go-site/tree/master/metadata/gorefs) for our [PLANA_REF reference pages](https://github.com/obophenotype/planaria-ontology/tree/master/metadata/planarefs)

And I followed the GO_REF format in the YAML.

### DIFF
--- a/metadata/db-xrefs.yaml
+++ b/metadata/db-xrefs.yaml
@@ -3200,3 +3200,14 @@
       url_syntax: http://zfin.org/cgi-bin/ZFIN_jump?record=[example_id]
       example_id: ZFIN:ZDB-GENE-990415-103
       example_url: http://zfin.org/cgi-bin/ZFIN_jump?record=ZDB-GENE-990415-103
+- database: PLANA_REF
+  name: Planaria Ontology Database References
+  generic_urls:
+    - http://purl.obolibrary.org/obo/plana/references/
+  entity_types:
+    - type_name: entity
+      type_id: BET:0000000
+      id_syntax: \d{7}
+      url_syntax: http://purl.obolibrary.org/obo/plana/references/[example_id]
+      example_id: "0000001"
+      example_url: http://purl.obolibrary.org/obo/plana/references/0000001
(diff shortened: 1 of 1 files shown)


cmungall:
looks like it doesn't resolve yet

For GO_REFs, we expand the PURL as follows:

https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/go.yml#L47-L51

Yes, it does maybe seem a bit insane having this double system, but they are separate concerns. You can of course just use your plana/references purls directly (this is more common for groups like you who work in OWL, Protege supports URLs as first class entities).

srobb1:
If we use " plana/references purls"  do I create a directory called references in the root of our repo and put our in our refs? Can I still use .md files or do they need to be just the \d{7}?

srobb1:
Ok. Making Progress

- purl.obolibrary.org/obo/plana/references/0000001
resolves to
https://raw.githubusercontent.com/obophenotype/planaria-ontology/master/references/0000001
**but it is not found.**


- purl.obolibrary.org/obo/plana/references/planaref-0000001.md
resolves to 
https://raw.githubusercontent.com/obophenotype/planaria-ontology/master/references/planaref-0000001.md
**and can be found**

- purl.obolibrary.org/obo/plana/references/planaref-0000001
resolves to
https://raw.githubusercontent.com/obophenotype/planaria-ontology/master/references/planaref-0000001
**and can be found**


In the Yaml can I modify the URL syntax like this?
      url_syntax: http://purl.obolibrary.org/obo/plana/references/planaref-[example_id].md

srobb1:
purl.obolibrary.org/obo/plana/references/0000001
resolves and is found now too.

kltm:
@cmungall Does this look right?

cmungall:
I think it's a bit suboptimal from plana's point of view, but at least it resolves to something, even if a bit opaque. We can work with OBO on providing a better general purpose solution for REFs in the future

PR #334, opened 2017-04-12
Clarify protein binding representation

cmungall:
@vanaukenk I can't seem to select you as a reviewer, can you or @monicacecilia take a look at this? @dougli1sqrd is implementing this rule and may have questions.

Note many of these rules were written pre-Noctua. For this one I have added a link to the github ticket https://github.com/geneontology/molecular_function_refactoring/issues/29

### DIFF
--- a/metadata/rules/gorule-0000004.md
+++ b/metadata/rules/gorule-0000004.md
@@ -18,7 +18,7 @@ to each MOD to evaluate and include the inferences in their GAF/DB.
 
 ## Representation of binding in the GO
 
-We use the term GO:0005515 to represent instances of protein binding. If a gene G is annotated to this term, its function involves binding another protein. The partner protein is represented in the with/from field of the association, i.e. the 'evidence' is the partner protein.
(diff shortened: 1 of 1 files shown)


vanaukenk:
Will do, Chris.

vanaukenk:
Hi @cmungall @dougli1sqrd and @monicacecilia 

My understanding of this annotation rule was that it was implemented to make sure there were reciprocal binding annotations for genes/gene products that are annotated to GO:0005515 or one of its children.  

To that end, I would explicitly state that in the rule doc; maybe something like this:

+We use the term GO:0005515 and its children to represent instances of protein binding. If a gene G is annotated to this term, its function involves binding another protein. The partner protein is represented in the with/from field of the association, i.e. the 'evidence' is the partner protein.  Annotation of gene G to a protein binding term requires that the partner protein also be annotated to a protein binding term, resulting in reciprocal protein binding annotations.  Note that the specific protein binding term used for annotation does not have to be the same for each partner.

+

+Note that this annotation rule predates the existence of annotation extensions (c16 in the GAF). It is more logical to specify the binding partner in c16, this would also allow cleaner separation of evidence from in-vivo activity, but for historic reasons the with/from field continues to be used.

cmungall:
ok, made @vanaukenk's changes, I'm going to merge. can do subsequent edits later

PR #338, opened 2017-04-25
WIP: Removing MGI doubling (DO NOT MERGE YET)

cmungall:
One proposal is for AGR to use this fork


cmungall:
Will need to be synced with https://github.com/identifiers-org/registry/issues/8

PR #344, opened 2017-05-11
Xenbase.yaml metadata file

malcolmfisher103:
Xenbase.yaml metadata file with details of public facing address for gzipped Xenbase.gpi file.


kltm:
@cmungall question about caps, and possibly the non-standard newlines.

cmungall:
Although the pipeline does not depend on it, I'd prefer lowercase for id and filename - easy to change at your end @malcolmfisher103? If you deal with the files we can edit the PR

If the newlines cause issues we'll fix.

malcolmfisher103:
No problem changing the case, I'll do the same for the GPI file as well. The newline thing is weird, for some reason they seem to have changed  to windows EOL format from linux at some point. I'll put a a corrected version with the right format.

The gpi file on our FTP site will take a little while to change it's name as that is run by our Calgary team.

malcolmfisher103:
Hopefully everything is fixed now. The GPI file on our FTP site has also been renamed and updated.

kltm:
@cmungall To clarify, this was originally started to get the identifiers into the pipeline to enable noctua annotation, yes?
As it stands now, the next time NEO is built (this should be automatically picked up?) and the index regenerated, input should be possible. However, we'll still need to do a minerva restart to get it to pickup the new NEO as well.

PR #345, opened 2017-05-11
ZFIN updates to users.yaml

doughowe:
Made a few updates to users.yaml for ZFIN people.  Removed md5s, added emails, GitHub user names, ORCIDs, and a couple new people.


kltm:
@doughowe 
Getting there!
You can see that GitHub is complaining of conflicts in your attempted PR (https://github.com/geneontology/go-site/pull/345/conflicts)--it looks like it doesn't like the way you moved your own name.

Also, just peering at the changes, I'd suggest that you remove the tag `email-md5` completely, rather than leaving it empty. For example, what you did for Yvonne instead of what you did for Leyla.

kltm:
Okay, closer. It now is "mergable", but it fails the schema testing: https://travis-ci.org/geneontology/go-site/builds/231317159?utm_source=github_status&utm_medium=notification

One is possibly a repeat of the ORCID 'http://orcid.org/0000-0002-2246-3722' ?

doughowe:
I don't see any duplicate uri.  I consolidated the ZFIN people all to one spot in the file ~ line 2000 for ease of editing.  How do I re-commence the testing?

kltm:
@doughowe With what you did, the code retested automatically. The error in question is:
```
ERROR: metadata/users.yaml:1973:8 [/225/uri] mapping key is duplicated.
```
I'm reading that as the `uri` dupe occurring on like 1973 now.

You should note that in this block, there is both `uri: 'GOC:sr'` and `uri: 'http://orcid.org/0000-0002-2246-3722'`. This is likely causing the problem.

doughowe:
@kltm 
After a few more updates...I think it's finally ready to go.

kltm:
@doughowe It all looks good to me. For future reference, we'll prefer the format given at the bottom of the file, i.e:
```yaml
 authorizations:
   noctua:
     go: 
       allow-edit: true
```
vs.
```yaml
 authorizations:
   noctua-go: 
       allow-edit: true
```
Of course, the file is mostly filled with the latter, so we'll have to port it with a macro, so it's not a big deal at all, but at some point the latter will be deprecated.

I cannot tell you how much we appreciate groups (and people) taking care of their own metadata--it's a huge help to us and gives you a lot more control. Thank you again for sticking through this. (And if you want to write-up a little how-to... :)

kltm:
@doughowe I've refresh the login system. The new metadata should now be live.

PR #347, opened 2017-05-15
Update paint.yaml

huaiyumi:
Changed the source of the paint GAF files to the PANTHER ftp site.

### DIFF
--- a/metadata/datasets/paint.yaml
+++ b/metadata/datasets/paint.yaml
@@ -11,7 +11,7 @@ datasets:
    dataset: paint_cgd
    submitter: paint
    compression: gzip
-   source: http://geneontology.org/gene-associations/submission/paint/pre-submission/gene_association.paint_cgd.gz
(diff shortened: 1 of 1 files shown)


cmungall:
thanks @huaiyumi 

 * I made edits to you branch to fix the URLs (you had e.g. paint_cgd.gz, when on your site it is paint_cgd.gaf.gz)
 * Previously there were some concerns about FTP continuing to be supported at USC - can you confirm that this is stable? Once it goes live it will cause issues if it breaks
 * I don't fully understand @selewis's objection - should we have a call to discuss this?

huaiyumi:
Thanks Chris for fixing the URLs. 
We will continue to support FTP. We are actively working with the USC IT group on this, so it should not be an issue.

cmungall:
Thanks. I chatted to @selewis and we're good with the changes and pulling from your ftp.

Is the code that generates the IBA's in github? I have a few comments and would like to add to the tracker.

PR #352, opened 2017-05-24
adding ontobio gaf parsing to instead of mikes script

dougli1sqrd:


dougli1sqrd:
Should we make a ticket for that in ontobio?

cmungall:
+1

dougli1sqrd:
Since ontobio is updated, we probably need another release. Then go-site can depend on the newest version with --filter-out and this can be merged

cmungall:
looks good

PR #353, opened 2017-05-24
Update db-xrefs.yaml

selewis:
Make NCBIGene primary and NCBI_Gene a synonym


cmungall:
@tonysawfordebi @kltm any objections?

kltm:
I've not really processed how long any catch-ups would take, but nothing in principle.

tonysawfordebi:
Fine with me.

PR #357, opened 2017-05-26
swap out rdfox for arachne

dougli1sqrd:
So this assumes we have `arachne` on the path. Not sure if that's the direction we want to go. Also, this PR only currently works with a change I made to arachne that's not released yet that only grabs .ttl files in the folder supplied by the --data option. So we should wait to merge this until Jim can look at my other PR, merge, and then release a new arachne.

### DIFF
--- a/pipeline/Makefile
+++ b/pipeline/Makefile
@@ -5,7 +5,7 @@ RDFOX_MEM ?= 32G
 GAF_OWL = target/go-gaf.owl
 ONT_MERGED = target/go-graphstore-merged.ttl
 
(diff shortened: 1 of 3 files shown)


kltm:
@cmungall I'd defer to you, but I would definitely prefer specific invocation over depending on the PATH. (I like the explicitness of the jars.)

dougli1sqrd:
It's true. It's how the other binaries are being executed right now, so I think I'll keep it as a PATH requirement. If we want, we can change how we're doing the others (owltools, etc) but I think I want everything to be done the same at least.

Issue #212, opened 2016-08-18
All the same databases / sources should have identical prefixes


kltm:
This is from @ValWood on https://github.com/geneontology/amigo/issues/298.

Her initial comment was:

```
This is probably really a jenkins QC check issue so it could be 'wrong tracker'
Shouldn't all the sources match the database prefix in some file somewhere?
...
```

The current "bad" list is:
- WormBase and WB
- FlyBase and FLYBASE
- UniProt and UniProtKB
- AspGD and ASPGD
- GO_Central and GO_CENTRAL and GOC (?)
- ENSEMBL and Ensembl
- REFGENOME and RefGenome

Possibly useful comments:
- As we migrate the perl gaf checks to jenkins, will this reporting become more transparent?
- It looks like somebody is filtering though an old db system and then we're getting them back.

kltm:
Noting this again, from a different context. Kind of worrying that a year has gone by on this.

Looking into this, it seems that at least some is due to upstream resources "capping" assigned_by, which we then ingest.

@cmungall

cmungall:
We should have the new GAF pipeline report and repair, assigning @dougli1sqrd

pgaudet:
Redundant with https://github.com/geneontology/go-site/issues/1218

PR #362, opened 2017-06-02
Add notes on identifiers for rule1

cmungall:
Also fixed URL for db-xrefs yaml


kltm:
@cmungall It might be good to add in the known information from https://github.com/geneontology/go-site/issues/212 ?
If we have a *should* mapping, we might want to make a YAML file with the suggested mapping somewhere.

cmungall:
We can certainly add known issues as an informative note.

As for the mapping, is that not trivially obtainable from db-xrefs.yaml?

kltm:
Yes, sorry--population of the synonyms field. I'll check that it is up to date.

PR #365, opened 2017-06-02
adding sparql for gorule 0000006

dougli1sqrd:
Paging @balhoff for advice also, but I can't add him as a reviewer


dougli1sqrd:
actually I think it might not. I need to double check this.

PR #366, opened 2017-06-02
sparql impl of gorule 7

dougli1sqrd:
Paging @balhoff for advice also, but I can't add him as a reviewer

### DIFF
--- a/metadata/rules/gorule-0000007.md
+++ b/metadata/rules/gorule-0000007.md
@@ -37,29 +37,31 @@ implementations:
       PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
       PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
       PREFIX owl: <http://www.w3.org/2002/07/owl#>
-      PREFIX ro: <http://purl.obolibrary.org/obo/RO_>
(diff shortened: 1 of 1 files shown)


cmungall:
minor: let's use uppercase 'GO' as the prefix. Semantically identical but good to be consistent.

balhoff:
@cmungall not sure how you feel about this but these can be even more readable by putting all the IDs in the prefixes:

```sparql
PREFIX CatalyticActivityMolecularFunction: <http://purl.obolibrary.org/obo/GO_0003824> 
# elided...
?j a CatalyticActivityMolecularFunction: . 
```

cmungall:
+1, I use this pattern a lot. And I usually separate with underscores 
and retain source capitalization, rather than CamelCasing

dougli1sqrd:
@balhoff @cmungall I can update the prefixes in this way. Should I change all the prefixes to be all caps?

balhoff:
I would not make it all caps in the cases where you're putting a whole readable name in. For example, both of these are good:

```sparql
PREFIX GO: <http://purl.obolibrary.org/obo/GO_>
PREFIX catalytic_activity: <http://purl.obolibrary.org/obo/GO_0003824>
```

For `GO:`, you would be appending the number in the query body, like `GO:0003824`. For `catalytic_activity:`, you would just use it as a placeholder by itself.

PR #367, opened 2017-06-02
adding sparql query for gorule 14

dougli1sqrd:
Paging @balhoff for help since he can't be a reviewer.


dougli1sqrd:
Does this look good?

PR #368, opened 2017-06-08
Graphstore rule checker

dougli1sqrd:
also paging @balhoff


dougli1sqrd:
I agree with both of these. I didn't know about prefix commons at the time. I agree about the later part also.

PR #380, opened 2017-06-23
Added groups to MGI curators-please check syntax

ukemi:


kltm:
Sorry, @ukemi . You seem to be copying a mistake that was injected recently by another users--there should be no trailing slash on any of the group URLs.
I.e. `http://geneontology.org/` -> `http://geneontology.org`

kltm:
Issue with slash here:
https://travis-ci.org/geneontology/go-site/builds/246337683#L176

PR #384, opened 2017-06-27
Update groups.yaml

srengel:
added SGD


kltm:
@srengel The text you added is not YAML. As well, there are upstream conflicts that would need to be resolved.

kltm:
@srengel Fixing while I'm here.

PR #406, opened 2017-07-14
updated Gene3D URL

mah11:


kltm:
Assuming @mah11 knows what's going on.

PR #407, opened 2017-07-14
added PROQUEST stanza for dissertation referencing

austinmeier:
Added DBX ref stanza for Proquest to point to thesis and dissertations.

### DIFF
--- a/metadata/db-xrefs.yaml
+++ b/metadata/db-xrefs.yaml
@@ -3227,3 +3227,13 @@
       url_syntax: http://purl.obolibrary.org/obo/plana/references/[example_id]
       example_id: "0000001"
       example_url: http://purl.obolibrary.org/obo/plana/references/0000001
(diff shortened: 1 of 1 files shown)


kltm:
Any comment here @cmungall ?

cmungall:
guess it's OK but in these cases I'd just use the URL directly in PO

kltm:
@austinmeier thoughts?

austinmeier:
Proquest holds records for other thesis and dissertations that might be helpful in the future.  I'm okay just using the URL as well. 

Request for this reference: https://github.com/Planteome/go-site-xrefs-fork/issues/1

austinmeier:
We will use the URL (until a need to add proquest database arises again!)

cooperl09:
We could use he URL in this case. If we get more refs from this source, we can revisit.

Issue #231, opened 2016-10-06
Move curator_group_contacts.tsv to yaml in go-site


cmungall:
We have two parallel redundant activities going on

@dosumis is doing this for the ontology group:
https://github.com/geneontology/go-ontology/blob/master/curator_group_contacts.tsv

meanwhile there is the ongoing activity by @kltm to organize group information for Noctua purposes: https://github.com/geneontology/noctua/issues/350

all metadata should go in go-site in yaml

kltm:
I'd just go ahead and use this as a seed for the next step in completing geneontology/noctua#350 ; does that sound about right?

kltm:
We could just close this out at point to it there. It should be completed shortly, and up until the final step, I'll seed from the TSV.

kltm:
Really, #106, besides the fact that we have the groups.yaml, #106 is really just things like this.

dosumis:
Any progress on this?  (Discussed on Ed call)

kltm:
@dosumis Well, if there are no more changes to this information for now, you could go ahead and help merge the TSV with the users.yaml--my information can be used as a model, and there is the users.schema.yaml file if there are any questions. It would be good to fix any inconsistent naming in there. We'll probably also want to move to just using the groups list in the future and deprecate the organization field (which means we'll also want to fill in the groups.yaml file).

kltm:
This should be completed with https://github.com/geneontology/go-site/pull/415

kltm:
I'm leaving the removal in go-ontology to @dosumis or others.

cmungall:
To summarize:

 * the tsv now lives in go-site/metadata as a csv https://github.com/geneontology/go-site/blob/master/metadata/group-contacts.csv
 * we are retaining fields from original (name, github). No email. No reqt so far
 * other metadata about groups lives in groups.yaml

This seemed easier than having a complex nested structure in groups.yaml

PR #412, opened 2017-07-19
Add contacts to groups schema

cmungall:


metadata/groups.schema.yaml
```diff
@@ -19,3 +19,26 @@ sequence:
         type: str
         required: true
         unique: true
+      "active_from":
+        type: str
+        required: false
+        pattern: /^\d\d\-\d\d\-\d\d\d\d$/
```
kltm: Would it be possible to get that reversed for iso 8601?

kltm:
It might be good to talk about changes like this.
For the schema, what is an `emaul`? Also:
https://github.com/geneontology/go-site/pull/412/files
Would it be possible to get the date reversed for iso 8601?
Also, for some types of sensitive information, we have go-internal; historically we've had email address count as sensitive.

kltm:
Also, is the intention to have this folded in as well?
https://github.com/geneontology/go-site/issues/231

metadata/groups.schema.yaml
cmungall: +1

cmungall:
yep, can go ahead and remove email (we only have gh in tsv so far) and fix dates iso was intent

PR #415, opened 2017-07-19
counter proposal to #412 and #413

kltm:
counter proposal to #412 and #413


kltm:
@cmungall 
In an ideal world, the groups in the CSV are, or an additional "column" is, the ID of the group. However, this is a hopefully at least a step in the right direction in getting the metadata out of go-ontology.

PR #417, opened 2017-07-21
python script invoked explicitely by python3

dougli1sqrd:


dougli1sqrd:
Hold on before merging. There's something not quite right.

cmungall:
Yes, wouldn't expect this to work:

```
python3 ontobio-parse-assocs.py
```

because the script isn't the first argument we don't invoke the shell's ability to use $PATH.

This would work:

```
python3 `which ontobio-parse-assocs.py`
```

but it would be very ugly to repeat. Maybe better to:

```
PARSE_ASSOCS := python3 `which ontobio-parse-assocs.py`

...

target:
    $(PARSE_ASSOCS) -i $< -o $@
```

cmungall:
overkill but also possible:

```
define wrap
    python3 `which $(1)`
endef

PARSE_ASSOCS = $(call wrap,ontobio-parse-assocs.py)
```

PR #420, opened 2017-07-24
Adding WB to groups.yaml

vanaukenk:


kltm:
@vanaukenk Just a reminder that you'll also have to add the groups that you belong to to the users.yaml if you want to be able to change "hats".

PR #422, opened 2017-08-14
new entry for iPMTnet

mah11:


kltm:
@mah11 I made some in-line changes to pass the checks--you had tabbing incorrect in your addition.

PR #424, opened 2017-08-24
using blazegraph-runner to load and reason gaf models into blazegraph

dougli1sqrd:


pipeline/Makefile
```diff
@@ -145,10 +142,15 @@ BG = java -server -XX:+UseG1GC -Xmx$(BGMEM) -cp $(BGJAR) com.bigdata.rdf.store.D
 # Load blazegraph
 
 BG_PORT = 8899
```
kltm: We should either have found clear port, or at least randomized port, used here.

dougli1sqrd: That's true. For the moment though this isn't being actively used. Once we have sparta checking integrated though, we should definitely think about that.

kltm: Okay, talked to @dougli1sqrd , all good

dougli1sqrd:
Something's wrong with the makefile, let's not merge yet.

dougli1sqrd:
Everything is fixed: https://build.berkeleybop.org/job/go-gaf-release-experimental/146/

PR #435, opened 2017-09-30
Update paint.yaml

huaiyumi:
The paint gaf files have been updated to reflect the current progress and the conversion to PANTHER v.12.0.


kltm:
Checked one, seems to download.

PR #439, opened 2017-10-04
Update zfin.yaml

sabrinatoro:
A gpi file is now available for ZFIN. I have added the information for this file.

### DIFF
--- a/metadata/datasets/zfin.yaml
+++ b/metadata/datasets/zfin.yaml
@@ -31,7 +31,6 @@ datasets:
    type: gpi
    dataset: zfin
    submitter: zfin
-   compression: gzip
    source: https://zfin.org/downloads/zfin.gpi
    entity_type: 
    status: active
--- a/metadata/datasets/zfin.yaml
+++ b/metadata/datasets/zfin.yaml
@@ -23,3 +23,19 @@ datasets:
    taxa:
     - NCBITaxon:7955
 
+ -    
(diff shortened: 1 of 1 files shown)


kltm:
@sabrinatoro 
I had a question about this. In the file, you have `compression: gzip`, but the source is not compressed, and there appears to not be a compressed file on the server. Is there a plan to gzip the GPI?

sabrinatoro:
@kltm : You are right, the source is not compressed (my mistake). 
At this point, there is no plan to gzpi the GPI . So the yaml file should be update to compression: none? (or should I add nothing? remove that line?). 

However, if it is better to have a gzip format, I can ask the devs to update the file.

kltm:
@sabrinatoro 
If there is no intention to compress, the line would just be removed completely. I believe that the compressed versions have been fairly standard in the case of GAFs, so I'd assume this tradition would have utility; if there is not much additional burden, it might be good to add.
Pinging @cmungall in case he wants to chime in here.

sabrinatoro:
OK, I will remove the line about the compression, and will add it back when we get a compressed file.

kltm:
Cheers!

cmungall:
Look ok now? (at top of this page look at "Files changed"), thanks!

kltm:
@sabrinatoro Is this closable now?

sabrinatoro:
yes, I think we can close. Thanks

PR #440, opened 2017-10-04
Update zfin.yaml

sabrinatoro:
removed the line for the gpi file:  compression: gzip 
(since the file in the source is not compressed)


cmungall:
I'm not sure I follow - https://zfin.org/downloads/gene_association.zfin.gz is compressed

cmungall:
OK, it looks like you intended to edit the GPI entry from #439 and edited the block about the gaf instead? Reopen if I'm not correct

PR #441, opened 2017-10-09
Added constraints for the new HTP-related evidence codes

tonysawfordebi:
Created rules for the following new codes:
ECO:0006056 (HTP), ECO:0007005 (HDA), ECO:0007001 (HMP), ECO:0007003 (HGI), and ECO:0007007 (HEP)

The new rules are essentially clones of the non-HTP equivalent codes


cmungall:
Thanks!

@dougli1sqrd we should make sure that these are propagated to gorules

PR #442, opened 2017-10-09
Added JSON-LD context for users and groups files.

balhoff:
cc @cmungall

### DIFF
--- a/metadata/users-groups-context.yaml
+++ b/metadata/users-groups-context.yaml
@@ -1,0 +1,22 @@
+rdfs: 'http://www.w3.org/2000/01/rdf-schema#'
+xsd: 'http://www.w3.org/2001/XMLSchema#'
+nickname: 'rdfs:label'
+label: 'rdfs:label'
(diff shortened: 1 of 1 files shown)


balhoff:
Ruby for converting to ttl, in case it is useful...

Dependencies:
```bash
gem install json-ld
gem install linkeddata
gem install rdf-turtle
```

Script:
```ruby
#!/usr/bin/env ruby

require 'json/ld'
require 'YAML'
require "rdf/turtle"

context_file = ARGV[0]
data_file = ARGV[1]
output_file = ARGV[2]

context = YAML.load_file(context_file)
data = YAML.load_file(data_file)

triples = JSON::LD::API.toRdf(data, {:expandContext => context})

RDF::Turtle::Writer.open(output_file) do |writer|
  triples.each do |statement|
    writer << statement
  end
end

```

cmungall:
Thanks.

In future we may want to use jena to avoid too many languages in the travis, after this: #238

PR #443, opened 2017-10-11
Add TAIR:Communication: to db-xrefs.yaml

elserj:
I'm not 100% sure I did this right, but I tried to follow what was done for TAIR:gene and TAIR:locus.  We found that we had a few reference links with TAIR:Communication and wanted them added to db-xrefs.yaml.  Should also point out that links in AmiGO seem to work without this added, but for other downstream projects thought this should be added.


kltm:
We'll see if it works in time. Passes the tests, so let's find out...

PR #444, opened 2017-10-16
Update users.yaml

vanaukenk:
Adding groups to WB curators.


kltm:
@vanaukenk 
How soon do you need these live?

vanaukenk:
@kltm - No rush on this.  Within the next few days is fine.

PR #445, opened 2017-10-20
File extensions

dougli1sqrd:


kltm:
@dougli1sqrd does this produce ttl and ttl.gz? Once produced, will blazegraph loading be able to operate just from ttl.gz?

dougli1sqrd:
@kltm yeah I had this keep the ttls around, cause that's what blazegraph and other targets use. Otherwise I have to do the zip/unzip business that we already do with the gaf artifacts.

kltm:
From the skype commentary, it looks like we might want to change blaze-runner to accept gzipped TTLs.

kltm:
From more skype commentary, and in-person convo, probably not right now.

PR #447, opened 2017-10-24
sparql implementation for gorule 8

dougli1sqrd:
This sparql query will find all classes that have the do_not_annotate properties on them. Then, any annotations or ?sub ?pred ?obj where the ?obj is one of the do_not_annotate classes represents a violation of this rule.


dougli1sqrd:
This logic isn't strictly correct though actually. We need to be able say that annotations to classes with gocheck_do_not_manually_annotate are allowed if the evidence for that annotation is an IEA.

PR #448, opened 2017-10-25
Update genedb.yaml and (finally) remove tsetse

kltm:
(finally) remove tsetse


kltm:
Going ahead with this we I can purge the buckets in preparation for tonight.

PR #467, opened 2017-11-27
Create gorule-0000024.md

pgaudet:
https://github.com/geneontology/go-annotation/issues/1707

### DIFF
--- a/metadata/rules/gorule-0000024.md
+++ b/metadata/rules/gorule-0000024.md
@@ -1,0 +1,9 @@
+---
+layout: rule
+id: GORULE:0000024
+title: "   "
+proposer: "pascale.gaudet@swiss.sib"
+status: Pending
+fail_mode: soft
+---
+Please see: https://github.com/geneontology/go-annotation/issues/1707


kltm:
@pgaudet Is this actually a PR or an new issue?

pgaudet:
It's a real issue!

kltm:
@pgaudet Is this PR intended to be a fix or a placeholder for the "actual" ticket?

kltm:
@pgaudet Nevermind--it seems that a lot of the rules in there are stubs anyways, so this would be normal?

kltm:
@pgaudet Nevermind--it seems that a lot of the rules in there are stubs anyways, so this would be normal?

pgaudet:
@kltm 
Yesterday on the ontology call @cmungall told us that this is where rules should be stored; isn't it the case ?

kltm:
From the strike-through above and the message after that, this seems to be a quirk of the rules that non-functional stubs are added. From the outside it looks rather odd is all.

pgaudet:
Hi @kltm I added this to the next ontology call; can you attend so we can have your input on what the SOP should be ?

kltm:
@pgaudet When is the next ontology call?

pgaudet:
Monday 9 am pacific

kltm:
Unfortunately, that conflicts with another call I have. I'll see if I can find a proxy.

cmungall:
None of the existing entries should be stubs.

Making stub PRs is OK, but they will not be merged until there is a textual description of the rule. Curators can make stubs if they want and assign a godev member to help complete. @pgaudet can you add more description on this one and we'll take it from there?

This PR should help clarify the process: #559

PR #470, opened 2017-11-29
Update group-contacts.csv

pgaudet:


kltm:
I assume that this is correct and will not disrupt any downstream processing.

PR #473, opened 2017-12-06
added zfin.gpi information in the ZFIN yaml

sabrinatoro:
The previous gpi information which I submitted in October was missing in this file (?).
I have added this information.
Note that compared to the previous version, we now have a compressed version of the file.


kltm:
I believe this is correct.
@cmungall I'm going to merge; please revert if I'm mistaken.

PR #478, opened 2017-12-07
Update users.yaml

vanaukenk:
Adding groups for Birgit, Livia, Helen, and Giulia.  Updating orcids for WB.

### DIFF
--- a/metadata/users.yaml
+++ b/metadata/users.yaml
@@ -429,6 +429,8 @@
       allow-write: true
     noctua-go:
       allow-edit: true
+  groups:
+    - 'http://flybase.org'
   accounts:
     github: hattrill
 -
@@ -737,6 +739,8 @@
   xref: 'GOC:bhm'
   uri: 'http://orcid.org/0000-0003-4062-6158'
   organization: InTact
+  groups:
+    - 'https://www.ebi.ac.uk/intact'
   comment: 'Birgit Meldal'
   accounts:
     github: bmeldal
@@ -1943,7 +1947,7 @@
   authorizations:
     noctua-go:
       allow-edit: true
-  uri: 'GOC:dr'
+  uri: 'https://orcid.org/0000-0002-4945-5837'
   xref: 'GOC:dr'
   organization: WB
   groups:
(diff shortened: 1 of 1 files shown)


kltm:
Hm, we now have both `http` and `https` URIs in the YAML. @cmungall Any thoughts on this?

vanaukenk:
Hi @kltm and @cmungall 
Any thoughts on the http vs https?
I can change the orcid entries back to http, if https is a problem.

kltm:
@vanaukenk Currently, all 150 are `http` and that's is what will be in all downstream files. My druthers at this point would be `http`.

kltm:
I can fix this inline if you'd don't have time today.

vanaukenk:
One thought, though.  The https actually came in via cut-and-paste from orcid.org, so if someone else populates that field via cut-and-paste, they may also wind up with https and maybe wouldn't know to fix it.  Does https cause problems elsewhere?

kltm:
Currently, they are different IDs in all downstream data.
If we update, we either have to update all models to date or accept that people will have different IDs for a while.
As we are a little short on time, I'd actually be fine with http and https mixed as long as nobody changes.

vanaukenk:
Ah, I see.  So if all my GO-CAM models currently use http and I update my URI to now be https, then will the attribution of my annotations to WB in the GPAD still work?  If not, then yes, we should return all the WB curator entries back to http.

kltm:
They will work, but there will be two different IDs, which may make reporting difficult.
For the moment, my suggested course of action would be: have stable IDs, regardless. For people who already have ORCIDs listed, they should not change to use `https`. For new people, either is fine as long as they do not change later on until we mass migrate everybody over to `https`.

kltm:
It looks like the only remaining changes would be the four group additions and adding a URI for GOC:dr. If that's the case, a new PR may be easier.

vanaukenk:
Okay; will do.  Thx for your input on this one.

kltm:
My apologies for the back and forth and trouble on this one while we worked out what makes the most sense over here.

Issue #469, opened 2017-11-28
Generate ttl versions of users.yaml and groups.yaml


balhoff:
A JSON-LD context is available for translating `users.yaml` and `groups.yaml` into RDF Turtle:

https://github.com/geneontology/go-site/blob/master/metadata/users-groups-context.yaml

Somehow, the GO pipeline needs to use the context to convert those to files to RDF, and make the RDF versions available online. Some general steps are required:

- Convert all 3 files from YAML to JSON.
- Use a JSON-LD processor to convert those to RDF using that context. 
  - Not all JSON-LD processors provide a way to specify an out-of-band context to be used in the translation. Thus, a preprocessing step may be needed to insert the context in the right place within the files. E.g. this might be needed for Jena `riot`.
  - I provided example Ruby code in #442 which can do the translation direct from the YAML files, but this adds dependencies to the pipeline.
- Publish the RDF files somewhere.

Data from these files is needed for use in Noctua/Minerva, e.g. https://github.com/geneontology/noctua/issues/502

balhoff:
A couple of notes if we want to do this with Jena `riot`:

- The `@nest` directive doesn't seem to be supported yet (it's not in the standard yet anyway)
  - It's not needed for the immediate use case, so that line could be commented out in the context.
- The data files have JSON arrays at the root, rather than objects. The root needs to be an object (`{...}`) in order to have a `@context` key. The data array can be put as the value for a `@graph` key.

The files can be processed to look like this:
```json
{
"@context": {
  "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
  "xsd": "http://www.w3.org/2001/XMLSchema#",
  etc.
  },
"@graph": [  existing contents  ]
}
```

A file like this with extension `.jsonld` can be opened directly by Jena tools.

kltm:
@balhoff @cmungall I assume the easiest way forward for the application of this would be to add the YAML->RDF to the metadata step of the build and then aim invocations at skyhook. This assumes that nobody else should be interested in our metadata as RDF, except incidentally in our downstream data products.

balhoff:
@kltm does the pipeline environment already have a command-line way to convert YAML to JSON?

kltm:
Um, yes. If you are plotting out how to do this, you can just assume that as a given.
Technically, it would work something like:
```
sjcarbon@moiraine:~/local/src/git/go-site[master]$:) ./node_modules/.bin/yaml2json --help
[...]

PR #488, opened 2017-12-18
Add script to convert YAML files to Turtle.

balhoff:
This script can be used to support production of users and groups ttl files for #469.


kltm:
@balhoff 
What is the `riot` command here?

balhoff:
Part of Jena command line tools.

balhoff:
I just pushed a fix for a misuse of the tmpfile variable.

kltm:
Okay, we'll have to figure out a way of getting Jena onto all the machines then. Either that, or add it to the "software" steps of the build.

balhoff:
Yep, either that or instead the ruby packages for the approach in https://github.com/geneontology/go-site/pull/442#issuecomment-335249178

PR #490, opened 2017-12-20
awk short circuit for goa_uniprot_all and make target

dougli1sqrd:


pipeline/util/goa_filter
```diff
@@ -0,0 +1,54 @@
...
+        # print "cache: is model organism"
+        next
+    } else if (taxa in goa_cache) {
+        # taxon is not a model organism, so we print it
+        # print "cache: is not a model organism"
+        print $0
+        next
+    } else {
+        # print "taxon not in caches, finding it json"
+        command="python util/model_organism.py --silent " taxa " ../metadata/datasets"
```
kltm: python, not python3 or path? I can see this might be difficult in a jenkins environment...

kltm: Would that interpreter run every time?

kltm:
I'm a little unsure about the python script in there; what about something like:
```
grep -h -e "[Tt]*axon:[0-9]*" metadata/datasets/*.yaml | sort | uniq | cut -d ":" -f 2
```

pipeline/util/goa_filter
cmungall: I had envisioned a one time compilation step to generate the awk or something similar

dougli1sqrd: No this isn't run every time. It saves off in an array taxons it has seen already, and checks there first. Really though, I think this could be simplified with little different approach. This was just my first pass. I was thinking about changing this so that there was a make target to find all the model organism taxa, and then we could load that file here one time at the start, and check against that. I think that is ultimately a cleaner and faster solution, I just hadn't gotten to that yet. That would also take care of the interpreter path issue. And you're right, jenkins could have a difficult time with that.

dougli1sqrd:
I saw this comment later. This just finds anything that is spelled like a taxon in the datasets I guess? Yeah I suppose we could do something like this instead of parsing the yaml. This seems dangerous. But maybe there's something we could do that's a little more structurally aware of the file?

kltm:
Assuming the npm utility yaml2json is available, something like this also works:
```
yaml2json metadata/datasets/fb.yaml | jq ".datasets[].taxa[]"
```

PR #493, opened 2017-12-21
Use ROBOT instead of riot for JSON-LD conversion.

balhoff:
Change JSON-LD conversion script to use ROBOT. This removes the dependency on Jena tools, as requested in https://github.com/geneontology/pipeline/issues/11.

### DIFF
--- a/scripts/yaml2turtle.sh
+++ b/scripts/yaml2turtle.sh
@@ -11,6 +11,7 @@ contextyaml=$1
 inputyaml=$2 
 outputttl=$3
 tmpfile=`mktemp`.jsonld
+echo "Using temp file: $tmpfile"
 echo '{"@context": ' >$tmpfile
 yaml2json $contextyaml >>$tmpfile
 echo ', "@graph": ' >>$tmpfile
--- a/scripts/yaml2turtle.sh
+++ b/scripts/yaml2turtle.sh
@@ -2,16 +2,18 @@
 
 # This script will convert a YAML file to Turtle format,
 # given a JSON-LD context, also in YAML format.
+# Since ROBOT is used, other file extensions for the output 
+# can select other formats in addition to Turtle.
 # Run: ./yaml2turtle.sh context.yaml data.yaml data.ttl
 
 contextyaml=$1
 # expected to have an array at the root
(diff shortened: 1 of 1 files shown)


scripts/yaml2turtle.sh
```diff
@@ -2,16 +2,18 @@
...
+# Since ROBOT is used, other file extensions for the output 
+# can select other formats in addition to Turtle.
 # Run: ./yaml2turtle.sh context.yaml data.yaml data.ttl
 
 contextyaml=$1
 # expected to have an array at the root
 inputyaml=$2 
 outputttl=$3
-tmpfile=`mktemp`
+tmpfile=`mktemp`.jsonld
```
kltm: Minor, but maybe: `mktemp -p .` to keep everything in directory for debugging?

balhoff: No such option on MacOS. We could switch to a fixed filename. I just didn't want to accidentally overwrite anyone's files, nor leave temporary stuff hanging around.

kltm: Fair enough--that would be a little hard to ensure at your end.
Can you use your own template? The reason I ask is that when we have multiple pipeline running and they are the tempfiles in /tmp it will be hard to distinguish between them. Ideally, they will have the tempfiles in their own workspace directory.

balhoff: Do you mean `mktemp -t yaml2turtle`, which will return something like `/var/folders/2r/4wk2vrj50xn2jk4wqbywws4w0000gq/T/yaml2turtle.pQsskKFc`?

kltm: Ugh--apparently the command is somewhat different on both systems: https://stackoverflow.com/questions/31396985/why-is-mktemp-on-os-x-broken-with-a-command-that-worked-on-linux#31397073
As we want full traceability without farfing around too much with cross-platform commands, could you just have it write/echo the name of tempfile to STDOUT or STDERR (e.g. "Using tempfile: foo")? That way we can at least look in the Jenkins logs to know which one was used.

balhoff: Okay, added a commit.

PR #494, opened 2017-12-21
splitting awk output into noiea and full versions

dougli1sqrd:


dougli1sqrd:
Oh yeah, that variable is old, I'll remove it. Do you want me to document in the awk script that it has the noiea side affect? Or in the make file?

cmungall:
both! :)

PR #499, opened 2018-01-04
Fix missing example ids and URLs for cribi_vitis, CO_125, OBO_SF2_PO …

elserj:
…and OBO_SF2_PECO


kltm:
@elserj Cheers! 
Line 408 for `CO_125` has:
```yaml
      example_url: url_example
```
I believe that is the final issue currently.

PR #500, opened 2018-01-08
Paint merged

dougli1sqrd:
So this has a script that merges the paint gaf zips (defined by the metadata) into the corresponding mod gaf zips. This produces a product `target/groups/{mod}/{mod_merged}.gaf.gz`. I produced a separate file just so that we can filter how we grab our data products.

### DIFF
--- a/metadata/datasets/paint.yaml
+++ b/metadata/datasets/paint.yaml
@@ -2,7 +2,7 @@ id: paint
 label: PAINT
 description: "GO data for PAINT"
 datasets:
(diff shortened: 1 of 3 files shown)


kltm:
@cmungall I think we talked about this before, but I'm a little uncomfortable with the `merge` id being different than the `id`, which is actually just the GAF name.

kltm:
@cmungall I'm still a little worried about this, but I'm merging just so we can move on.

PR #501, opened 2018-01-08
hopefully this wont summon cthulu, but this changes the makemaker to …

dougli1sqrd:
…generate targets for noiea goa_uniprot_all

### DIFF
--- a/pipeline/util/generate-makefile.py
+++ b/pipeline/util/generate-makefile.py
@@ -11,7 +11,7 @@ import yaml
 from json import dumps
 
 SKIP = ["goa_pdb", "goa_uniprot_gcrp"]
-ONLY_GAF = ["goa_uniprot_all"]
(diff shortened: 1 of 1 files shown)


kltm:
I suspect this will either break or fall through the sanity checks, but it's probably to get this worked out early in the month.

dougli1sqrd:
Oh maybe. It seems like it's a super set of what we were doing before. As in, we just do extra processing on the noiea, and other downstream stuff won't know the difference? Maybe? haha, Whatever I can do to help.

kltm:
Discussed with @dougli1sqrd . He'll do some testing with the sanity scripts and get a version working with the `iteration` branch.

dougli1sqrd:
Looking over your sanity script, I anticipate fewer problems than maybe you thought? But only experiment will show...

PR #502, opened 2018-01-08
added noiea target to goa rule. model_organism script fixed to includ…

dougli1sqrd:
…e goa taxa

### DIFF
--- a/pipeline/util/model_organism.py
+++ b/pipeline/util/model_organism.py
@@ -13,7 +13,7 @@ def cli():
 @click.option("--out", "-o", type=click.File("w"))
 def taxons(dataset_dir, out):
     datasets_paths = [os.path.abspath(os.path.join(dataset_dir, dataset)) for dataset in os.listdir(dataset_dir) if dataset.endswith("yaml")]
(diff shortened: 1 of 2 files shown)


pipeline/util/model_organism.py
```diff
@@ -24,9 +24,8 @@ def read_taxa_for_path(dataset_path):
     with open(dataset_path) as group_file:
         group_data = yaml.load(group_file)
         gaf_data = [data for data in group_data["datasets"] if data["type"] == "gaf"]
-        taxa = [ taxon for gaf in gaf_data if gaf["taxa"] for taxon in gaf["taxa"] ]
+        taxa = [ taxon for gaf in gaf_data if gaf["taxa"] for taxon in gaf["taxa"] if taxon.split(":")[1] != "1" ]
```
kltm: I do not believe a special case is needed. But probably won't harm anything.

kltm:
@dougli1sqrd Is `click` a standard library in py3?

dougli1sqrd:
It's not, but I've been using it in our other pipeline tools, and you agreed at some point that click could be installed on our pipeline machines.

kltm:
Looking at what we have, it's within the pipeline section that already has the requirements.txt run, so all good.

PR #503, opened 2018-01-10
paint_merge.py now checks if merge_into key exists in metadata

dougli1sqrd:


pipeline/util/paint_merge.py
```diff
@@ -14,7 +14,10 @@ def cli():
 def merge(paint_metadata, groups_dir):
     metapaint = yaml.load(paint_metadata)
     merger_groups = metapaint["datasets"]
-    merge_data = { dataset_id_to_path(dataset["id"], groups_dir): merges_into_path(dataset["merges_into"], groups_dir) for dataset in merger_groups if dataset["merges_into"] != "other" }
+    merge_data = {
+        dataset_id_to_path(dataset["id"], groups_dir): merges_into_path(dataset["merges_into"], groups_dir)
+            for dataset in merger_groups if "merges_into" in dataset and dataset["merges_into"] != "other"
```
kltm: "other" is not necessarily needed anymore, yes?

dougli1sqrd: I suppose not. It just worries me that since that's how we used to do it, someone might write metadata like that again in the future.

kltm: But they would be wrong and there should be breakage, right?
IIRC, there is no "other" GAF to merge into anyways.

PR #507, opened 2018-01-16
Update users.yaml

kltm:
Add @dustine32


kltm:
Travis did not pick up test, so just hot merged.

PR #513, opened 2018-01-18
using make maker to make merged paint sources

dougli1sqrd:


kltm:
Talked to @dougli1sqrd , let's test it out.

PR #531, opened 2018-02-14
Update groups.yaml

cmungall:
Removing spaces from labels

The YAML label field should probably eventually map to something other than rdfs:label, as here it has a more specialized meaning for a space-free code for a provider

See https://github.com/geneontology/noctua/issues/539


kltm:
This is meant to be a user readable label *only*. If this is being overloaded with something else, I think we need to talk about it.

PR #537, opened 2018-02-15
add shorthand for gaf compat

kltm:
added info for geneontology/noctua#539


kltm:
Not sure why travis is not checking this, but will merge anyways...

balhoff:
They're experiencing degraded service right now.

PR #547, opened 2018-02-28
renaming directory indexer with underscore for module loading purpose…

dougli1sqrd:
…s, added up one level,


kltm:
Looking good and functional.

PR #549, opened 2018-03-01
upload now uses a function that chooses multipart if the file is too big

dougli1sqrd:


kltm:
apparently all good to go now

PR #572, opened 2018-03-20
Adding SynGO to groups

dustine32:
Needed for newly generated SynGO Noctua models.


dustine32:
Sorry @kltm got caught in a meeting. Yes, I was wondering which way to go too but it seems like an umbrella term was desired: "I propose to add "SynGO" as a more suitable label for grouping all annotations provided by our consortium members." @ftwkoopmans , does this sound right?

kltm:
Fair enough. I'll just go ahead and merge now.

ftwkoopmans:
yea that's what I had in mind, thanks !

PR #585, opened 2018-03-28
updating groups from new users sanity check

dougli1sqrd:


kltm:
@dougli1sqrd For the changes from `http` to `https`, has that data already gone live and is being used? If so, that would be another cleansing step that would be required.

metadata/groups.yaml
```diff
@@ -68,15 +68,23 @@
...
-  label: 'FlyBase'
+  label: 'SPOT (EBI)'
+  id: http://www.ebi.ac.uk/spot
+  shorthand: 'SPOT'
+-
+  label: EMBL-EBI
+  id: http://www.ebi.ac.uk
+  shorthand: EBI
+-
+  label: 'FB'
```
kltm: Why this distinction? `FlyBase` would be the more readable version...

dougli1sqrd: Chris said that the `FlyBase` needed to be the shorthand, as that matches the needed field in GAF. But many of the users have `FB` as their group. So I wanted `FB` to be the label in order to detect that when a user says `FB` they mean `FlyBase`.

dougli1sqrd:
Oh, I thought you had said that `https` was fine as long as it resolved? I can turn them all to `http` though. And no, this groups is not being used anywhere yet. I'd have to update the users file too though. But as we haven't committed that yet, that's not a problem.

metadata/groups.yaml
kltm: Sorry, I'm not understanding why not just having the label be "FlyBase" too?

kltm:
`https` is indeed fine, but only for new resources. For established resources, we'd have to do a data clean-up as technically they are different things.

metadata/groups.yaml
dougli1sqrd: Because people had created their user organization as `FB`. I would have had to change people's organization from `FB` to `FlyBase` when making the repair script. But if a user's org should not be `FB` we could change that their in users.yaml.

dougli1sqrd:
ah I see, great. I hadn't realized I changed the urls, I can fix those.

PR #592, opened 2018-04-06
Update wb.yaml

vanaukenk:
Hi,
I've updated the wb groups.yaml to include the latest WB gpi file.
Can @dougli1sqrd and/or @kltm please review this to make sure everything looks okay?
Thx.


kltm:
@vanaukenk I see no reason to think that wouldn't be fine.

PR #596, opened 2018-04-10
Update goref-0000114.md

bmeldal:


bmeldal:
GH is doing something strange about the + preceding every attribute adding a ++ so Tony suggested to remove them.

PR #609, opened 2018-04-14
Update db-xrefs.yaml

cmungall:
Fixing MESH expansion, fixes https://github.com/EBISPOT/OLS/issues/204

Note: we should transition to using existing resolvers where we don't add any value ourselves. But not sure if this causes confusion for systems like OLS that do a mashup taking ours as canonical and mixing in with 3rd party resolver, e.g. id.org or n2t, cc @SimonJupp

### DIFF
--- a/metadata/db-xrefs.yaml
+++ b/metadata/db-xrefs.yaml
@@ -1698,15 +1698,18 @@
       example_id: MEROPS_fam:M18
       example_url: http://merops.sanger.ac.uk/cgi-bin/famsum?family=m18
 - database: MeSH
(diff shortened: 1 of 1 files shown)


kltm:
It might be good to look at the difference between URIs and community resolution. While we have a set of places to go for more information about entities we care about, those might not be the same in other community contexts. I'd be fine if we we could do periodic downloads of the preferred mapping for our community from a more centralized location (I do like offline access), but experience suggests that what we have now works really well for updating, tracing, and control.

Issue #453, opened 2017-11-08
All Noctua curators should have a group associated in the YAML file


pgaudet:

pgaudet:
Note: default group is the first in the list.

vanaukenk:
Added to agenda for 2017-11-14 annotation call:

http://wiki.geneontology.org/index.php/Annotation_Conf._Call_2017-11-14

kltm:
To clarify this a bit more, as policy moving forward, all curators need to have group information (at least one) specified in their `users.yaml`. This group also needs to be specified exactly the same way in the `groups.yaml`. As it stands now, the curator group information can be gappy.

Ideally, there is a contact person for each group that can help triage their own data in these files. For future use, it may be good to add the relevant metadata in the `comments` of the `users.yaml` and/or create a group for these people in GH.

kltm:
As well, as an early workaround before we settled on ORCIDs, we allowed people to use either a URI or to fall back onto their GOC xrefs. I believe we should stop this practice and require all curators to have an identifying URI (very very preferably an ORCID). Any xref they have must be kept around so we can do a one-time forward mapping.

kltm:
Also want to tag on @cmungall for that last bit.

cmungall:
groups actually used in models:

```
http://dictybase.org
http://geneontology.org
http://informatics.jax.org
http://lbl.gov
http://purl.obolibrary.org/go/groups/BHF-UCL
http://purl.obolibrary.org/go/groups/IntAct
http://purl.obolibrary.org/go/groups/UniProt
http://purl.obolibrary.org/go/groups/WB
http://www.ucl.ac.uk/functional-gene-annotation/cardiovascular
http://www.ucl.ac.uk/functional-gene-annotation/neurological/projects/tabs/aruk-ucl
http://www.wormbase.org
http://www.yeastgenome.org
http://zfin.org
https://www.nextprot.org
https://www.ucl.ac.uk/functional-gene-annotation/neurological/projects/tabs/syngo-ucl
```

kltm:
@cmungall Closure would be that all past and current curators have correct metadata.
https://github.com/geneontology/go-site/issues/453#issuecomment-343007547
It may be that this can be confirmed and has been done.

kltm:
@vanaukenk @pgaudet 
We are currently finishing up updating users and groups, patching up as best we can. This is the current list of users we are planning to drop Noctua edit permissions for:
```
[...]

PR #615, opened 2018-04-17
Fix users.yaml with respect to noctua permissions

kltm:
This PR:

* adds travis tests for users and groups with respect to noctua permissions
* tries to normalize the use of org
* adds `previous-groups` list option
* adds groups where possible, extracted from org
* removes permissions from those that do not have ORCIDs

With this, org can be deprecated

Issue #453


kltm:
Pinging @cmungall -- I just want a quick second set of eye on some rather large changes.

PR #625, opened 2018-04-23
 Update users.yaml for SynGO ORCIDs

dustine32:


kltm:
@dustine32 Could you add a newline to the bottom.

As well, consider that the `organization` field is now optional and pretty much deprecated. Ideally, these users would have groups (as specified in the groups.yaml file) which could then be used in the pipeline and other processes. It's fine for a user to have more than one group.

dustine32:
Thanks! For sure, I'll commit again with a newline.

For the org field, for now I'll blank these out. When I have time, I can try to update groups.yaml  adding the orgs that Frank sent, then add these new guys into the appropriate users' groups after in addition to the "SynGO" group. Does this sound like a plan?

kltm:
@dustine32 That sounds good. However, in the interim, there is no harm in keeping the `organization` as-is.

dustine32:
Cool! Should be updated now. I left the orgs as-is. It took me 40 minutes to add a newline.

PR #652, opened 2018-05-19
Pass exclusions on

dougli1sqrd:
This uses some additional operating on the datasets to exclude anything in the EXCLUDES variable from being run as a make target from the metadata yamls.

### DIFF
--- a/pipeline/Makefile
+++ b/pipeline/Makefile
@@ -3,15 +3,24 @@ clean:
 
 METADATA_DIR=../metadata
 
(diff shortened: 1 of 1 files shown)


dougli1sqrd:
https://build.geneontology.org/job/geneontology/job/pipeline/job/master/270/ Passes. In order for this to get into the midnight build, I'm going to go ahead and merge this.

kltm:
@dougli1sqrd As the control variables get more mixed, might these not be better pushed into the YAML files at some point?

PR #657, opened 2018-05-23
Update Makefile

kltm:
We are gunna hot commit.


kltm:
@dougli1sqrd Says this is fine.

PR #661, opened 2018-05-24
Splitting paint

dougli1sqrd:
This moves all the paint dataset yaml into a new field in the standard group dataset yaml. Now for some dataset in a group <d>, if there is was a paint source paint_<d>, now the metadata for paint_<g> will be in dataset entry. 

The datasets schema has been updated to reflect this.

There needs to be a corresponding ontobio change to read the metadata files, which exists here: https://github.com/biolink/ontobio/pull/180

There is a test run in Jenkins here: https://build.geneontology.org/job/geneontology/job/pipeline/job/splitting_paint_iteration/2/

Lastly, the order of operations for this is to merge ontobio, release 1.1.0, and then merge this PR. This should be done right after another though, so they remain in sync.


cmungall:
Thanks! Is there a way to see the files produced by Jenkins?

We will need to send out announcements about the change in the metadata. We have only just managed to get people trained in how things work at the moment.

In future can we separate out big changes like this into their own PR, and have a dedicated issue to discuss large scale changes?

cmungall:
I assume this can be closed?

PR #665, opened 2018-05-30
Update GO annotations for L. major and T. brucei

Achchuthan:
Source URLs have been changed for these two organisms and you can pull the data for these from the respective ftp links. The files are gzipped and provided in the ftp site

### DIFF
--- a/metadata/datasets/genedb.yaml
+++ b/metadata/datasets/genedb.yaml
@@ -16,7 +16,7 @@ datasets:
    dataset: genedb_lmajor
    submitter: genedb
    compression: gzip
-   source: http://geneontology.org/gene-associations/submission/gene_association.GeneDB_Lmajor.gz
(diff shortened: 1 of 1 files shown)


kltm:
`curl ftp://ftp.sanger.ac.uk/pub/project/pathogens/as44/Tbruceibrucei927.gaf.gz | gunzip - | less`
Seems similar to the original, but with a gaf 1.0 header.
`curl ftp://ftp.sanger.ac.uk/pub/project/pathogens/as44/Lmajor.gaf.gz | gunzip - | less` seems to produce nothing/is empty, which is a change from the previous file.
@Achchuthan is this correct?

Achchuthan:
Dear @kltm,

Sorry for the inconvenience. If you please try to download from the same link again, L. major file should have data now. Both T. brucei and L. major files should be similar to the original version except that there are more rows of data now and GeneDB scripts produce gaf 1.0 header.

Thanks,
Achchuthan

kltm:
Let's see how it goes!

Issue #642, opened 2018-05-12
Remove paint as top level product in pipeline, add goa_chicken to iteration branch


dougli1sqrd:
Paint should not be a top level product. Other mods and groups that need paint will build paint themselves already. Including paint now is just redundant, and it's extremely slow.

Also, goa_chicken caused an ontobio hiccup, so add this dataset to the goa yaml in the iteration branch for testing.

selewis:
Nor am I following this. The first impression is that this is just wrong.
The mods need something to consume to incorporate at their resources. But
we are never going to be consuming anything PAINT related that is received
from them, ever, ever, ever...

dougli1sqrd:
So, the pipeline has in the past produce paint as a top level data product. But the gaf pipeline flow is slightly different now. 
Now we have:
[dataset]-src.gaf --> [dataset]_valid.gaf --> [dataset].gpi

If there exists an entry for paint_[dataset].gaf in the datasets yaml, we do:
paint_[dataset]-src.gaf --> paint_[dataset].gaf (using [dataset].gpi to correct gene information)

Then we perform the merge:
[dataset]_valid.gaf + paint_[dataset].gaf --> [dataset].gaf (The final gaf product).

Now, this process produces a paint_[dataset]_valid.gaf, a paint_[dataset]_noiea.gaf, and reports for the parsing. 

If we treat paint as it's own product, we produce mostly redundant gaf products, since the main datasets will already produce the paint gaf products. So, removing paint as a top level product just removes what is mostly a redundant pipeline execution.

But yes, we can definitely loop back next week about the specifics, or chat this weekend.

cmungall:
This is fine. The paint_foo.gaf will end up in the products folder, not the main release folder

selewis:
Should we perhaps talk to Huaiyu about having PAINT export GPAD? It'll have
to happen sooner or later.

kltm:
As an internal note, talking a little with @dougli1sqrd, there is currently some non-uniformity in the treatment of inputs to the final merged file, causing things like `*_valid.gaf.gz` to exist in some cases, but not others, and for `paint_*.gaf.gz` to be severely reduced from `paint_*_valid.gaf.gz`. Part of this refactor will be to have a more uniform set of files and their expected contents; to rationalize the evolution of what is going on with the files.

PR #668, opened 2018-05-31
Explain where PAINT annotations are located.

pgaudet:


kltm:
Also note that this will seriously change with https://github.com/geneontology/go-site/issues/642

cmungall:
This metadata file is currently used by the GO Central pipeline. Without it, the pipeline has no way of knowing where to fetch PAINT files from.

pgaudet:
Are you talking about the files that the 'Pipeline' loads, or how MODs should pick up their PAINT data ? This ticket was about the latter; I dont know if our response addresses that.

kltm:
@pgaudet This PR would shut down a critical part of the pipeline line. Unless this PR contains an alternate to the current pipeline, this cannot be accepted.

kltm:
@pgaudet "...This ticket was about the latter..."
You are not in the ticket/issue section, this is the PR (pull request) area and the change that you are trying to commit would severely adversely affect the current running of the pipeline.

pgaudet:
ok good thanks !

PR #680, opened 2018-06-07
Update rules.schema.yml

pgaudet:


dougli1sqrd:
I believe this currently exists now in master already. @pgaudet can you double check that this is done already and is in master? I think it went in with our big changes from a week or so ago

dougli1sqrd:
I'm closing this, because these changes are already present. If I'm wrong feel free to reopen and let me know.

PR #683, opened 2018-06-12
Gorule update 670

dougli1sqrd:
This is a work in progress presently. 

This is an effort to clarify the GO Rules. Some rules have been split up, some updated, and in general the hope is to get all the rules all on the same page.


dougli1sqrd:
So we'll merge this for now, and then continue to refine and make these better as we go along.

PR #689, opened 2018-06-29
documentation for the blazegraph public triple store

dougli1sqrd:


graphstore/triplestore_info.md
```diff
@@ -0,0 +1,138 @@
...
+The Blazegraph triple store gets produced as one of the steps of the pipeline.
+After the GAF annotations have been validated by [Ontobio](https://github.com/biolink/ontobio)
+the validated GAFs are loosely converted into GO-CAMs. There is roughly
+one "Annoton" per GAF annotation line. Once we have RDF GO-CAM of the GAFs,
+they are imported into the triple store. The specifics of how the this is
+organized in the triple store is discussed below.
+
+### Contents
+
+The Ontolgoy and all imports, all GAFs as GO-CAMs, and all production Noctua
```
kltm: sp

kltm:
@goodb would also likely be interested.

dougli1sqrd:
@goodb and @ukemi don't show up on the list of possible reviewers. But I'd love their reviews.

graphstore/triplestore_info.md
dougli1sqrd: No see, Ontolgoy is the joy of working on Ontologies.

kltm:
Added @ukemi and @goodb

graphstore/triplestore_info.md
```diff
@@ -0,0 +1,138 @@
...
+
+This allows easy querying between the asserted and inferred triples.
+
+In Noctua Model graphs, we add the triple
+
+    <graph_uri> <http://model.geneontology.org/graphType> <http://model.geneontology.org/noctuaCam> .
+
+This also applies to inferred Noctua Model graphs.
+
+In RDF converted
```
kltm: Is this a hanging statement?

ukemi:
This is exactly the kind of documentation we need. I still think the language can be cleaned up in places. We may want to expand a bit more on the background of some of the terminology if we want the whole GO community and not just those of us who have interacted with you closely to understand it completely.

goodb:
Who is the target audience for this?  Text like:

> After the GAF annotations have been validated by Ontobio the validated GAFs are loosely converted into GO-CAMs. There is roughly one "Annoton" per GAF annotation line

will be completely incomprehensible to people that are not deeply enmeshed in the project.  I think this doc is a fine start for developers within the GOC to start getting their hands on the RDF.  If you want people from the outside world to make use of the triple store I'd suggest building a specific document for them that focuses exclusively on the questions that they can answer with the store and how those questions can be posed in SPARQL.  Perhaps this is in the works to go along with the query examples implementation (SNORQL or whatever has been selected).  All of the stuff about how its built (pipeline/reasoning/etc.) can be moved out to developer docs to clarify the message for the users.

dougli1sqrd:
@goodb Yeah this is mostly intended for the go-ontology developers I think. Maybe @cmungall has more specific thoughts on who this is for. But you're right, this is sort of the beginning doc for a set of canned queries that we'll use an updated snorql for in the future. Also, this particular document is firstly also just putting this info on a page. Whether it stays here in go-site in this directory or as markdown is subject to change.

dougli1sqrd:
@ukemi I'd love specific suggestions on language. You can place comments directly in the document on lines that you find difficult or that can be improved. 

Also be sure to view this rendered version of the markdown also:
https://github.com/geneontology/go-site/blob/graphstore-documentation/graphstore/triplestore_info.md

PR #691, opened 2018-07-03
Updated file set & source locations

tonysawfordebi:
As a temporary measure, until dictyBase have finalised the updates to their database, GOA are generating a GPAD, GPI, and GAF file on their behalf. These files are located on the GOA ftp site, and are updated weekly.

### DIFF
--- a/metadata/datasets/dictybase.yaml
+++ b/metadata/datasets/dictybase.yaml
@@ -7,17 +7,53 @@ project_url: "http://dictybase.org"
 funding_source: "NIH (NIGMS and NHGRI)"
 email_report: "dictybase@northwestern.edu"
 datasets:
- -    
+ -
+   id: dictybase.gpa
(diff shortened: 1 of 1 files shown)


kltm:
@dougli1sqrd Looking at the PR from @tonysawfordebi , my instinct says that there could possibly be issues; I'd like you to take a look at it if you could.
The `source` is changing from:
  `http://geneontology.org/gene-associations/submission/gene_association.dictyBase.gz`
to:
  `ftp://ftp.ebi.ac.uk/pub/contrib/goa/dictyBase.gaf.gz`
Does the addition of the `.gaf` pre-suffix change anything in what we're doing? Possibly a hardwired id somewhere?

As well, I assume everything will be "ignored" at this point except for the `type: gaf`, correct?

kltm:
Talked to @dougli1sqrd ; there should be no effects.

PR #692, opened 2018-07-03
Update gorule-0000031.md

pgaudet:
Updated code to text


metadata/rules/gorule-0000031.md
```diff
@@ -11,18 +11,9 @@ implementations:
     source: https://github.com/biolink/ontobio/blob/master/ontobio/io/assocparser.py#L390
 ---
 ## In GAFs, relation is overloaded into qualifier.
```
dougli1sqrd: My only comment is we should probably take out the `##` in this line and in the following lines. These produce headings in markdown and when rendered this file looks pretty strange.

PR #693, opened 2018-07-03
Update users.yaml

vanaukenk:
Adding Marie-Claire Harrison, a new WormBase curator.
@kltm - when you get a chance can you update Noctua so Marie-Claire can start making annotations?
Thx.


kltm:
@vanaukenk I believe you all should be good.

PR #696, opened 2018-07-09
changing rule 8 to a report rule and fail soft

dougli1sqrd:
And this should really definitely fix the pipeline


kltm:
Is there a kwalify for the rules file?

kltm:
A good conversation here about what we'd want https://github.com/Grokzen/pykwalify/issues/35 , but the short answer is "no".

PR #733, opened 2018-07-31
Added example

pgaudet:


pgaudet:
@dougli1sqrd 
OK, but is there any way GORULE:0000014 can appear in the report ??

dougli1sqrd:
GORULE 14 is obsoleted in favor of using GORULE 20. We had done this after this PR I think. So no, gorule-14 no longer can appear in the report.

pgaudet:
> GORULE 14 is obsoleted in favor of using GORULE 20. We had done this after this PR I think. So no, gorule-14 no longer can appear in the report.

Sure - I meant in principle -

PR #743, opened 2018-08-02
Proposal for GOCAM CURIE

lpalbou:
I would rather use "GOCAM" than "gomodel", but since we are using gomodel:xxx everywhere, it's probably better for the moment. @kltm your opinion ?


kltm:
@lpalbou In retrospect, if we had had "GOCAM:" from the beginning, that may have been ideal. Unfortunately, that was many years ago. At least it's not "LEGO:", that would have been awkward... :)

Issue #716, opened 2018-07-30
Convert gorule-0000011 from legacy (owltools) to implemented 


pgaudet:
* ND annotations to root nodes only
* This should be a filter (annotations should be excluded) 

Thanks, Pascale

dougli1sqrd:
https://github.com/geneontology/go-site/pull/745

dougli1sqrd:
Oh, this was supposed to be a filter, okay. I can do that too.

pgaudet:
Also, we will remove the existing SQL, reword the rule to exclude the GOREF references, and update the sparql.

pgaudet:
(Note that GORULE_11 used to also make a restriction on the reference allowed, this will be moved to GO_RULE33)

pgaudet:
OK done

PR #745, opened 2018-08-03
adding sparql for gorule 11

dougli1sqrd:


dougli1sqrd:
https://github.com/geneontology/go-site/issues/716

dougli1sqrd:
I'll also add the ontobio filter for this too.

dougli1sqrd:
Yep that's true. You won't get "annotations" as they've been split into GO-CAM models. I can try and put maybe a little more identifying information, but it'll make the query moderately more complicated. I thought this would  be enough to serve as a way for people to find out more about what is in error. (I know the graph labels are dumb right now, but that's going to be fixed soon.) But it's a jumping off point. Besides, we'll have the ontobio check as well since we in fact want to filter these. But I'll see if I can make the query have a little more info.

pgaudet:
@dougli1sqrd I am not following. 
The documentation should represent what we are doing, so this query is not executing the rule described, so it cannot be defined as in **implementation** of the rule. 
I also dont know what you mean by 
> "annotations" have been split into GO-CAM models

Do you mean to say that in the process of loading GAF files from external sources those are converted to GO-CAMs? If that's the case, this should be described in the pipeline documentation (and we should also think about QC on that process). 

Thanks, Pascale

pgaudet:
Also, for me that query gives an error in Blazegraph. 
java.util.concurrent.ExecutionException: org.openrdf.query.MalformedQueryException: Lexical error at line 1, column 28. Encountered: "," (44), after : "gene_product.symbol" 

which is strange because I would assume you test these queries before dong a commit + PR ? Otherwise I recommend you do that in the future.

dougli1sqrd:
Sorry, I didn't see that there was a GO_REF:0000015 requirement. Is that what you are referring to as far as this not implementing the rule? Otherwise, this query will pick up annotations with `ND` that are to go terms other than the root ontology terms. The updated query will show you what gene product URI and what GO term URI this is. 

As far as the error in blazegraph, it looks like you pasted in the old deleted SQL in blazegraph. You see the error is complaining about `gene_product.symbol` on line 1? That is represented in the SQL query, but is not in the SPARQL. 

And yes, I write the queries in a SPARQL editor live against the RDF triplestore.

dougli1sqrd:
So in playing with this query, it has become apparent that there are over 100,000 violations of this rule, as written. I spoke with @cmungall about just the huge number of failures and he suggests that we write the rule to relax the GO_REF requirement. Perhaps if the GO_REF is incorrect we merely log a message, but we don't filter the line. Perhaps we only filter annotations of ND to non root terms? That's just a suggested solution, but 100,000 seems like a lot to filter. What do you think @pgaudet?

pgaudet:
@dougli1sqrd 
Where your rule is too strict of that you dont take into account the 'synonyms' of GO_refs, see http://www.geneontology.org/cgi-bin/references.cgi (I don't know where is the current version of this file)

GO_REF:0000015 also has External accession(s): AspGD_REF:ASPL0000111607, CGD_REF:CAL0125086, dictyBase_REF:2, dictyBase_REF:9851, FB:FBrf0159398, MGI:MGI:2156816, RGD:1598407, SGD_REF:S000069584, TAIR:Communication:1345790, ZFIN:ZDB-PUB-031118-1, GO_REF:nd 

Can you allow these?


Thanks, Pascale

pgaudet:
@dougli1sqrd I ran the query in blazegraph and the first line I get is 

  | <http://identifiers.org/uniprot/Q81Q41> | <http://purl.obolibrary.org/obo/GO_0003674> | <http://purl.obolibrary.org/obo/go/references/0000015>

This is allowed ! 

Looks like we are getting the correct annotations rather than the incorrect ones.

dougli1sqrd:
Yeah I know, I realized that yesterday, and I'm fixing it. Thanks for pointing it out though!

dougli1sqrd:
I updated the sparql yesterday. What do you think now of the results?

pgaudet:
@dougli1sqrd @cmungall I dont think we can 'relax' the rule (if that means allowing any reference), the reason being that people may read a paper that says 'this protein is uncharacterized' or 'the localization is unknown, etc, and curators may misinterpret this to mean ND in *this paper*. This is why we had implemented that reference in the first place. 

@vanaukenk please correct me if this is wrong, but I think the reference is important. 

Thanks, Pascale

cmungall:
How about this:

at parse time we should repair MOD pub IDs to be the canonical GO_REF (one day with AGR we will get to the stage where duplicate pub IDs are never minted in the first place). This is better for our users. And it will simplify the queries.

PR #757, opened 2018-08-09
use random port for sparts checks

dougli1sqrd:


kltm:
While not "perfect" (i.e. testing before use), this at least allows multiple runs with port clobbering.

kltm:
@dougli1sqrd 
A command like `nc -z 127.0.0.1 9909` can be used as a test. May as well die early and not *attempt* to reuse a port--at least that gives us a traceable cause of failure.

Issue #765, opened 2018-08-14
Change TAIR gene ID format in db-xrefs.yaml to match PAINT with values


dustine32:
Related to @tonysawfordebi 's error reporting [here](https://github.com/geneontology/go-annotation/issues/2044#issuecomment-411663480), which thinks TAIR ID's in with/from columns of IBA annotations are not using correct format. Ex: TAIR:locus:2117758. They seem to use the locus-based identifiers for genes as stated [here](https://www.arabidopsis.org/portals/nomenclature/guidelines.jsp#locus). I couldn't find any annotation results from Amigo by searching "TAIR:gene*" (currently the format for gene in db-xrefs.yaml) but I could for "TAIR:locus*".

I'm proposing that we update the `entity_type: gene` properties to match those in `entity_type: locus`:
```
- type_name: gene
  type_id: SO:0000704
  id_syntax: locus:[0-9]{7}
  url_syntax: http://arabidopsis.org/servlets/TairObject?accession=[example_id]
  example_id: TAIR:locus:2146653
  example_url: http://arabidopsis.org/servlets/TairObject?accession=locus:2146653
```

Tagging @thomaspd @huaiyumi since I discussed this with them. I'll also put in a PR with the changes.

dustine32:
Also, tagging this issue: https://github.com/pantherdb/fullgo_paint_update/issues/6

kltm:
https://github.com/geneontology/go-site/pull/766#issuecomment-413052864

kltm:
Noting https://github.com/geneontology/go-site/issues/659

kltm:
I'm worried that this looks like a pass at https://github.com/geneontology/go-site/issues/659, but conflating locus IDs with genes.

dustine32:
Yeah, I saw https://github.com/geneontology/go-site/issues/659 too. I can email TAIR helpdesk to confirm the `locus:####` format is how they mint all of their IDs. Or is email too square?

dustine32:
Me too! Emailed...

kltm:
Talking with @cmungall , it may be best to attack this in the eco constraints file (as you identified) with
```yaml
entity_type: *primary_transcript
```

dustine32:
@kltm Cool! I'll update the eco-usage-constraints.yaml then let you know when the PR's ready to review. Thanks!

tberardini:
See my comment in #659 please.

pgaudet:
[...]

PR #766, opened 2018-08-14
Updating db-xrefs.yaml for issue #765 - fix TAIR:locus ID validation …

dustine32:
…errors

### DIFF
--- a/metadata/eco-usage-constraints.yaml
+++ b/metadata/eco-usage-constraints.yaml
@@ -47,6 +47,9 @@ entity_types:
   - &rna
       id: CHEBI:33697
       name: ribonucleic acid
+  - &primary_transcript
+      id: SO:0000185
+      name: primary transcript
 
 eco_usage_constraints:
   - eco_id: ECO:0000353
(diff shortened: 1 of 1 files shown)


kltm:
@dustine32 Could you hold off on this for a minute...

kltm:
Well, this should break any of our stuff, as it's dumber. @tonysawfordebi Does this work for you and your smarter stuff?

tonysawfordebi:
@kltm Sorry for the delayed reply - was away last week

The proposed changes to eco-usage-constraints.yaml look good to me, so I say ship it...

tonysawfordebi:
I had forgotten one thing...

We need one more update in order to make all our validation stuff work correctly, and that is to add this entry for SO:0000185 into go-upper.obo:

[Term]
id: SO:0000185 ! primary_transcript
name: primary_transcript
is_a: SO:0000673 ! transcript

(see https://www.ebi.ac.uk/ols/ontologies/so/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FSO_0000185)

There is already an entry for SO:0000673 in go-upper.obo

tonysawfordebi:
I've updated go-upper.obo
(https://github.com/geneontology/go-ontology/pull/16284)

tonysawfordebi:
@kltm Any idea how long it will take for the changes that I made yesterday to find their way into http://snapshot.geneontology.org/ontology/extensions/go-upper.obo?

tonysawfordebi:
Thanks - it looks like everything's in place now.

dustine32:
Thanks @tonysawfordebi for fixing this!

PR #768, opened 2018-08-15
Updated broken link

pgaudet:


kltm:
What is go-dbxrefs.yaml?

kltm:
I believe we fixed this in @selewis 's commit a minute ago.

PR #797, opened 2018-09-04
fixing how rule validation works so rule error will be caught by travis

dougli1sqrd:
This will make sure that a validation error occurs when there is no yaml in the yamldown, and is just straight text.


dougli1sqrd:
FIxes https://github.com/geneontology/go-site/issues/798

kltm:
Looks clear to me.

PR #803, opened 2018-09-10
Removed URL

pgaudet:
since they don't exist.


kltm:
I think that these may be needed in the future; they are currently placeholders.
We'll probably have something like: http://current.geneontology.org/annotations/cgd.gaf.gz

kltm:
Needs to be redone with #804 in mind
https://github.com/geneontology/go-site/pull/803

PR #805, opened 2018-09-12
Issue 804 url field

kltm:
Updates to metadata, schema, and download page generator.


kltm:
This seems to have worked with a local doctored combined.reports.json file.

PR #806, opened 2018-09-12
add pombase gp2protein location

mah11:


kltm:
@dougli1sqrd Could you take a look at this? I believe the `url` field is erroneous, but I don't know how this would be otherwise processed.

mah11:
Sorry, my mistake - that URL should have been changed, I think to http://geneontology.org/gp2protein/gp2protein.pombase.gz

(I copied the GAF stanza and edited almost everything I should have ...)

kltm:
@mah11 Okay, let me see if I can clarify the situation. Mid-term, we'll be putting SVN into full legacy mode, which means no longer committing new files to it and no longer making its products available from SVN. As a step in that direction, for our primary data pipeline, we put out the call to update the metadata so that SVN was not mentioned (essentially http://geneontology.org/gene-associations/submission/), thinking about the GAFs. However, as you're doing here, there are other products that have been put into SVN that we were not currently thinking about and do not currently play a part in the pipeline, like the gp2protein file.
So, for our use case, we do not actually need to have this captured, especially as we would prefer to get this type of information via GPI (as you mention). Also, we're not sure who the current users of the gp2protein file at its current location are, so we're not sure of the consequences of not having it available.
@tonysawfordebi Would you happen to know anybody who is currently using the gp2protein files?

Also tagging @cmungall in case I forgot anything.

kltm:
@dougli1sqrd Approved.

PR #809, opened 2018-09-13
overall gorule reports page first pass

dougli1sqrd:


kltm:
Just a question, is click in the runtime environment for that section of the pipeline? In general, as nice as click is, a couple more lines means we wouldn't have to bring it in (if it's not already there)--my code as an example.

PR #817, opened 2018-09-24
Update sgd.yaml

suzialeksander:
GAF not updated yet (will be soon).  SGD uses the GPI and GPAD from EBI; are these changes the way to show this?

### DIFF
--- a/metadata/datasets/sgd.yaml
+++ b/metadata/datasets/sgd.yaml
@@ -26,7 +26,6 @@ datasets:
     - NCBITaxon:41870
     - NCBITaxon:4932
     - NCBITaxon:559292
-    
  -
    id: sgd.gpad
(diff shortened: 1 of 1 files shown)


kltm:
@suzialeksander I'm not quite following. The metadata in this file is exclusively for the ingest portion of our pipeline. Currently, the SGD upstream is through the legacy SVN (something that we're trying to get people to update): http://geneontology.org/gene-associations/submission/gene_association.sgd.gz
We do not take GPAD (GPA/GPI) files as input yet. That said, does SGD pull all annotations directly from P2GO?

dougli1sqrd:
This looks great, @suzialeksander . We just want to make sure that you'll update the GAF metadata when you get the new source URL in the upcoming future.

PR #822, opened 2018-09-25
first pass at creator of .arbre files from set upstream PANTHER files…

kltm:
…; work on geneontology/amigo#532


kltm:
Taking as just a new script--independent.

PR #823, opened 2018-09-26
goa_uniprot_all awk filter for IBA

dougli1sqrd:
https://github.com/geneontology/pipeline/issues/60


kltm:
Let's see how this goes.

Issue #785, opened 2018-08-28
All pipeline reports should have a date


dougli1sqrd:
All pipeline reports should have a date.

pgaudet:
@dougli1sqrd @kltm 
In fact, all files produced should have a date (the ontology, the annotations) 
Also, the ontology header should indicate whether it is a 'release' or a snapshot (right now snapshots have 'data-version: releases/2018-09-09' (for example). This is misleading. 

Should I create a separate ticket ? 

Thanks, Pascale

kltm:
@pgaudet Yes, please go ahead and create a separate ticket for any desired files that do not yet have dates.

PR #824, opened 2018-09-26
date added to report and a title to gorule report page

dougli1sqrd:
This should complete this: https://github.com/geneontology/go-site/issues/785


kltm:
@dougli1sqrd Actually, you'll need to read this as either a command line input argument (preferred) or as an environmental variable.
All displayed dates will need to be locked, no matter when individual parts run.

dougli1sqrd:
@kltm updated. Corresponding PR at Jenkinsfile here: https://github.com/geneontology/pipeline/pull/62

kltm:
Tis guud.

PR #828, opened 2018-09-27
Update gorule-0000031.md

pgaudet:


dougli1sqrd:
Did you mean to fix this rule instead? https://github.com/geneontology/go-site/blob/master/metadata/rules/gorule-0000032.md

This seems like a perfectly fine rule, no?

PR #830, opened 2018-09-27
Created new rule

pgaudet:


cmungall:
The title of this PR says "Created new rule", but it also deletes a rule too. Is this intentional?

pgaudet:
Probably not ! I'll start over.

PR #831, opened 2018-09-27
Created new GO rule 34

pgaudet:


dougli1sqrd:
I don't think we need this since we have gorule-14 and gorule-27, right? I guess the thing to do is to make a ticket saying ontobio should make sure and flag with/from as having obsolete IDs. But this seems encompassed by the previous rules.

pgaudet:
Hi @dougli1sqrd, you're right, this is probably not needed, but the other rules must be clarified. 

**-> gorule-0000027:** 
    Col 1 and all DB abbreviations must be in db-xrefs.yaml (see below)
    All GO IDs must be extant in current ontology
**-> gorule-0000014:** 
This rule is now merged with GORULE:0000020.
**-> GORULE:0000020**
Automatic repair of annotations to merged or obsoleted terms


(Pascale): this does not say whether it only looks in Col5 only or also in Col8  - I couldn't find that by looking at the script. 

Can you please double check ? 

We do need to reformulate (at least) one of the rule to make sure this is clear. 

Thanks, Pascale

dougli1sqrd:
https://github.com/geneontology/go-site/pull/840
https://github.com/biolink/ontobio/issues/253

I think with these things we can close this PR.

dougli1sqrd:
If the above proposed changes would satisfactorily subsume this new rule and PR, let's close this.

pgaudet:
@dougli1sqrd ok to close. 

@dougli1sqrd  @cmungall 
It's confusing to have these 2 repos for rules. Can we leave an open ticket here while things get implemented in onto-bio?

PR #834, opened 2018-09-28
set gramene reduction threshold to .3

dougli1sqrd:


kltm:
Surely *that's* low enough, right?

PR #839, opened 2018-10-02
gorule report page spiffification

dougli1sqrd:
The attachment is not actually a txt. Rename as html and view with browser to see current status.

[wb-report.txt](https://github.com/geneontology/go-site/files/2439253/wb-report.txt)


kltm:
Looking good!

PR #852, opened 2018-10-10
goa_chicken_complex allowed to be reduced by 70%

dougli1sqrd:


kltm:
Apparently playing a game of chicken.

PR #855, opened 2018-10-12
adding many groups and some dbs

dougli1sqrd:


dougli1sqrd:
So the Alzheimers University_of_Toronto website has an apostrophe in it. Which is url encoded as seen. This doesn't pass the url regex, so we'll have to fix that.

kltm:
I'm not sure we shouldn't try and fix that upstream. Having a quote seems to be absolutely *begging* to have issues later on--I'd rather we not bobbytables our ingest.

kltm:
Sorry, I misread that; yes, we need to update.

PR #858, opened 2018-10-12
update PomBase urls; fixes #857

### LINKED ISSUE #857: update the yaml for POmBAse links
amigo links 
from
https://www.pombase.org/spombe/
to
https://www.pombase.org/

I think this probably the dbxrefs file but I'm never sure where this is?

@mah11

mah11:


kltm:
Looks good.

PR #859, opened 2018-10-12
Update rgd.yaml

tutajm:
As of Oct 12, 2018, we will provide the current (weekly) file of rat GO annotations in GAF format at RGD github site. 

The file http://geneontology.org/gene-associations/submission/gene_association.rgd.gz will not be updated any longer.

### DIFF
--- a/metadata/datasets/rgd.yaml
+++ b/metadata/datasets/rgd.yaml
@@ -16,7 +16,7 @@ datasets:
    dataset: rgd
    submitter: rgd
    compression: gzip
-   source: http://geneontology.org/gene-associations/submission/gene_association.rgd.gz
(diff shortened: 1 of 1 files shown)


kltm:
@tutajm Unfortunately, this is not a file, but rather a web page, and cannot be used as  a replacement.
```
curl https://github.com/rat-genome-database/rgd-annotation-files/blob/master/gene_association.rgd.gz | grep -c "\<div\ class\="
```

tutajm:
Could you please slightly update the url (just append '?raw=true'):

https://github.com/rat-genome-database/rgd-annotation-files/blob/master/gene_association.rgd.gz?raw=true

kltm:
Perhaps https://github.com/rat-genome-database/rgd-annotation-files/raw/master/gene_association.rgd.gz might be better?

kltm:
If you're okay with that, we'll merge.

PR #863, opened 2018-10-17
Update sgd.yaml

suzialeksander:
Making sure I did this right? Pretty sure it's ok.


dougli1sqrd:
Looks good! Looks like a gaf when I download it!

PR #869, opened 2018-10-19
GOC as a seperate group. Lets verify the id URI

dougli1sqrd:


pgaudet:
Why are we changing this ? Don't other groups link to the top-level page of their resource ?

dougli1sqrd:
We're changing this so that GAFs that have GOC as a `assigned_by` won't trigger the rule 27. For now, we will just add GOC as a group instead of seeing 10,000 messages that GOC isn't in groups.yaml, and people can fix their files over time.

PR #874, opened 2018-11-01
removing sparql for rule 6 as it takes to long and we can implement i…

dougli1sqrd:
…n ontobio. Removing sparql for rule 11 as it already has an ontobio implementation


dougli1sqrd:
@kltm approves

Issue #699, opened 2018-07-11
Fix gorule-0000008.md


pgaudet:
Currently both subsets ( gocheck_do_not_annotate and gocheck_do_not_manually_annotate)
 are treated the same way; but it should be:
* gocheck_do_not_annotate  -> never allowed
* gocheck_do_not_manually_annotate -> only allowed for IEA

@ValWood @vanaukenk Is this correct?

ValWood:
Well historically InterPro asked to be able to make annotations to "do not annotate" terms which  is why we have the exception.

Sometimes, they may have a protein family where they can't be specific. I cant think of an example right now.
But this would be correct.

How many  "gocheck_do_not_manually_annotate" do we have?

ValWood:
I think most of these could be made "do not annotate"

Probably 
'SWI/SNF superfamily-type complex' 

but most others I don't see any reason why Interpro etc should not be subjected to the same, more stringent guidelines.
Especially for instance 'biological phase' GO:0044848  (this entire branch is only avaiable for use in annotation extensions).

v

ValWood:
Is it possible to find out if there are any InterPro mapping direct to these terms, and if so see if they can be made more specific?

pgaudet:
Maybe someone from interpro can tell us? Where is the IPR2GO mapping file?
Just compare?

Le mer. 18 juil. 2018 à 6:54 PM, Val Wood <notifications@github.com> a
écrit :

ValWood:
GO:0008152	metabolic process
maybe we  could relax this and allow   for IEA, but it isn't very useful. YOu will only usually have it for  high level EC classificaitns where you can't say anything also. I would say that for these it is better just to have the MF annotation (oxidoreductase, transferase etc), and not bother with a biological process (because we don't really know anything about the physiological role. To illustrate 3230/4659 pombe proteins with a process annotation are annotated to metabolic process. 

GO:0007049	cell cycle
We could relax this.  But most will be "mitotic cell cycle" or even more specific

GO:0006950	response to stress
I don' think this is a useful term for process. I would like it to be disallowed completely

GO:0000910	cytokinesis
>99% of these will be "mitotic cytokinesis"  . I have 602 papers about cytokinesis, and only 1 about meiotic cytokinesis. Meiotic cytokinesis is very different at least in fungi where the spindle pole body  nucleates the forespore membrane. I don't  really even know how it happens...


GO:0042710	biofilm formation
I don't really understand this term. It's more of a phenotype.... 

[...]

PR #876, opened 2018-11-01
taking out sparql impl, deferring to ontobio python impl, issue 699

dougli1sqrd:
For https://github.com/geneontology/go-site/issues/699


kltm:
Apparently the good kind of empty.

PR #888, opened 2018-11-08
Create gorule-0000036

pgaudet:


kltm:
@pgaudet 
```yaml
    enum: [implemented, proposed, approved, deprecated, legacy]
```

kltm:
Talking to @dougli1sqrd , this needs a little more specification. Just to be clear, we would want to "report annotations the involve gene products where the gene product is annotated to a term 'x' and 'regulation of X' (multiple annotations involved)".

Maybe this would be effective as a Val-style co-annotation rule, or have the implementation initially there as it is a "global" check?

kltm:
@cmungall This may be another to add to the list-y pattern that crops up a lot.

Also, is this specific to 'x' and 'regulation of x', or is it more general with all term pairs x and y, where y regulates x? Or to restate, is this a cleanse of "regulation of" terms?

kltm:
Okay.

PR #892, opened 2018-11-08
Created new gorule-0000037

pgaudet:


kltm:
I'm a little confused by this--is this meant to be "GO_Central", or is there to be a bulk upgrade of annotations that get through (via GO Rule 26) to be "upgraded" to assigned_by "PAINT"?

Will need a description to be a GO Rule.

pgaudet:
Well since we can make primary annotations using the group GO_Central, I dont think we can formally use that as the rule (ie "GO_Central should be the only source of IBA is not strict enough").

cmungall:
How about:

If the evidence code is IBA, then (1) the assigned_by field must be GO_Central (2) the reference field must be PMID:21873635 and (3) the annotation must have been sourced from the PAINT inference pipeline. 

Implementation: the GO Central pipeline should filter out IBAs from any submission source that is not a PAINT submission source, i.e. one registered in paint.yaml and having a name paint_X. Note that IBAs from PAINT will be injected in to the final release file for that organism.

(we could also make this more explicit by having a field in the yaml that indicates this is allowed, but I think hardcoding the above is fine)

pgaudet:
> The idea is that 0000026 and 0000037 should be complimentary, with 0000026 being the "uninteresting IBA" rule and 0000037 being the "interesting" rule.

Right - 0000026 is a filter of correct annotations that we want to ingest from another source- 
0000037  is a 'rule' in the sense that annotations are incorrect. 

Thanks ! Pascale

PR #893, opened 2018-11-08
Created new gorule-0000038

pgaudet:


kltm:
Talked to @dougli1sqrd and this will need a more clear/concise description to allow implementation. E.g., how is a sequence identified, etc. As well, would this be "global" or specific to a resource or species?

pgaudet:
Hi @kltm 
I think this is what P2GO checks do: 
For annotations by ISO, ISA or ISS, the annotations of the Gene product (col 8 info in the GAF) should be checked to ensure that an annotation exists to the same or a more granular term. For instance: 

Gene Product 1 | GO:1233456 |  ISS/ISO/ISA | with: Gene Product 2
Gene Product 2 | GO:1233456 or a child | EXP or a descendant

@tonysawfordebi Is this right ? 
@kltm @dougli1sqrd How would you like me to reformulate this ?

PR #896, opened 2018-11-08
Created new gorule-0000039

pgaudet:


kltm:
How is the protein complex identified in this case?

Rule needs a description for implementation.

kltm:
C2 is the local identifier, how would that be identified? Is this to check the namespace in c1?

pgaudet:
@kltm Yes 

What is the namespace for the Complex Portal ? (where do we store these?)

pgaudet:
Is this now accurate ?

Issue #897, opened 2018-11-08
New gorule0000040 Check for valid combination of evidence code and GO term


pgaudet:
P2GO check 

@tonysawfordebi 

Can you give the specific details of that check (which ECO for which terms? 

Thanks, Pascale

pgaudet:
https://github.com/geneontology/go-site/pull/898

tonysawfordebi:
P2G has three places where it does this check:

- evidence is IEP (ECO:0000270) or HEP (ECO:0007007) and GO term is not a process term
- GO term is one of the three root terms (GO:0005575, GO:0008150, GO:0003674) and evidence is not ND (ECO:0000307)
- evidence is ND and the GO term is not one of the root terms

pgaudet:
> evidence is IEP (ECO:0000270) or HEP (ECO:0007007) and GO term is not a process term
-> gorule-0000006

>  GO term is one of the three root terms (GO:0005575, GO:0008150, GO:0003674) and evidence is not ND (ECO:0000307)
-> gorule-0000011

>  evidence is ND and the GO term is not one of the root terms
-> gorule-0000011

So it looks like we don't need that rule.

PR #898, opened 2018-11-08
Created gorule-0000040

pgaudet:


kltm:
@cmungall So far in this GO Rules list, we've seen:

* gp/term constraints
* eco/term constraints
* eco/go_ref constraints
* maybe others I missed?

As there seems to be a more general pattern here, it might be a good idea to look at these and maybe have a single YAML (a la Tony's eco-constraints file) or something in the ontologies to do the lifting?

kltm:
Otherwise, we need a location and description of the file, as well as a description in the rule for implementors.

pgaudet:
Included in existing rules, see https://github.com/geneontology/go-site/issues/897

PR #900, opened 2018-11-08
Created new gorule-0000041

pgaudet:


kltm:
This appears to be the same as GO Rule 11, which can be reworded as one-to-one and onto.

pgaudet:
No, it's the reverse of Rule 11. I think we need both.

dougli1sqrd:
Why can we not have gorule-11 say something like:
```
annotations to the GO roots can only have ND evidence, and ND evidence can only used with annotations to the Go roots.
```

Like, this is a one to one mapping between the two ideas, being the ND evidence code and the GO roots.

pgaudet:
OK

PR #903, opened 2018-11-08
Created new gorule-0000042

pgaudet:


kltm:
The status should not be `implemented`.
Talked to @dougli1sqrd and this seems clear. This still needs a description.

PR #907, opened 2018-11-08
Created gorule-0000044

pgaudet:


kltm:
Talked to @dougli1sqrd and this seems fine. Would need to have a description, possibly use current title.

pgaudet:
Are descriptions really mandatory ? Seems weird if they are just a repeat.

dougli1sqrd:
Yeah, descriptions are required for a complete and full rule. There should be text describing exactly what's going on, in at least as much detail for someone to write a sparql implementation or some code. But a biological context could be nice too, just to document why this rule exists. It'd be like a GO term without a text definition.

PR #909, opened 2018-11-08
Created new gorule-0000045

pgaudet:


kltm:
@cmungall Is this intended/planned to be folded into the ontology, or is this something that would need to be an addition into ontobio more directly?

pgaudet:
We need to look at what the yaml file is checking:
https://github.com/geneontology/go-site/blob/master/metadata/eco-usage-constraints.schema.yaml

And check if this is how we want to encode them.

cmungall:
go ahead and merge

PR #911, opened 2018-11-08
Created new gorule-0000046

pgaudet:


kltm:
Talked with @dougli1sqrd and this will need more of a description before it can be implemented. To clarify, is this about GO:0043621? So, if has said term in closure, with/from must not be empty and must be topic protein?

pgaudet:
@tonysawfordebi Which terms do you test for this in P2GO ? 
GO:0043621 is 'self-association', meaning associating with the same molecule; it seems wrong to have two entities in this case - I would even say GO:0043621  should NOT have anything in the 'with' field.

RLovering:
But as we are exporting data from the with field to PSICQUIC then it might be useful for people to see self-binding annotations in their networks, plus the evidence for selfassociation is IPI and with themselves.
Ruth

pgaudet:
Let's implement this rule the same way as P2GO. I guess that means GO:0043621 is also included. 

@tonysawfordebi @alexsign Is this right ?

alexsign:
@pgaudet  Tony is away this week, he can clarify it when he is back next week. But, from what I can see from protein2go check is if the GO term is GO:0043621 (protein self-association), GO:0042803 (protein homodimerization activity) or GO:0042802 (identical protein binding), then standalone IPI annotations are allowed

pgaudet:
'Standalone' meaning without 'with' info ?

alexsign:
No, in this case, as I understand it, you don't have to create Reciprocal Annotation

pgaudet:
OK, thanks

alexsign:
I'm looking at the protein2go code now. I find another rule which basically: if it's self-binding term annotation, with_string and entity must be the same. If I find more rules related I'll post it here as well.

pgaudet:
@alexsign I suppose the same rule applies to IGI ?

alexsign:
@pgaudet  I don't see the same rules implemented in the code for IGI.

PR #913, opened 2018-11-08
Create gorule-0000047

pgaudet:


kltm:
Talked with @dougli1sqrd and this seems to be fine. Needs a description to be taken as a rule. Possibly use current title?

PR #915, opened 2018-11-08
Created new gorule-0000048

pgaudet:


kltm:
Talked with @dougli1sqrd and this seems to be fine. That said, the title should be moved to the description and a new title supplied.

PR #917, opened 2018-11-08
Create gorule-0000049.md

pgaudet:


kltm:
This rule appears to possibly be a little under specified? If this meant to be per resource or per species? Otherwise, it seems like it might be a somewhat random rule.

pgaudet:
Is it better now ?

PR #920, opened 2018-11-08
Created gorule-0000050

pgaudet:


kltm:
@cmungall Is this proposal still in the air? Has this progressed and/or will some of it be moved into the ontology? If so, we should probably have some more conception of what the blacklist actually is and how it will work within the system. (Would assume YAML into ontobio, etc., etc.)

cmungall:
There was a ticket somewhere. It can just be a simple 3 or 4 column CSV. gene/term/reason/URL-or-pub

kltm:
Okay, so to get this in, we still need to finish the prose/full description. My understanding is that it should be complete enough that somebody else can pick it up and work on the implementation?

So something like:
```
ontobio reads in CSV from command line flag and uses it to filter GAFs during parse. The CSV lives in XYZ. CSV must pass travis checks at ABC.
```

cmungall:
My idea for the rule specifications is that they are independent of the implementation or file format. the same rule can be implemented in sparql for dynamic checks in Noctua.

The general pattern:

 * normative specification in terms of datamodel
 * guidance for implementation in context A
 * guidance for implementation in context B

E.g.

_Do not annotate exceptions are encoded as tuples of (gene, term, reason, pmid). This indicates that gene/product should never be annotated directly to term, or to an ancestor of term following is-a and part-of. Any such annotations are invalid and should be blocked from entry into GO Central. No replacements are specified_

_For GAF or GPAD filtering, these are treated as errors and filtered. For implementation in a curation tool such as Noctua or PAINT the curator should be blocked from making such an annotation_

kltm:
It seems that the described failmode and the metadata fail_mode are different.

As well, to start implementation, it would be good to have a seed blacklist file to start with, but that's maybe not necessary while still `proposed`. 

@cmungall I think we need to have a discussion at some point about expanding the rules schema to support status for multiple implementations, as well to make searching for what a client implementer should look at when starting out.

PR #922, opened 2018-11-08
phrasing rule to say root terms should only use ND and ND only goes t…

dougli1sqrd:
…o root terms

For PR #900


pgaudet:
Is the rule implemented to check it both ways ?

dougli1sqrd:
Not currently, but it would be easy to change.

pgaudet:
OK, made the new ticket.

kltm:
See #926

PR #931, opened 2018-11-15
Updating title from proposed rule 37 and refering to rule 37 in the description

dougli1sqrd:


dougli1sqrd:
@kltm says it's okay

PR #946, opened 2018-11-29
Update users.yaml

tberardini:
Added Erica Bakker, TAIR curator.


tberardini:
@kltm  Can you review and integrate my changes, if they look alright?  Or assign to someone else, if that's more appropriate.  Thanks.

kltm:
Should be good.

PR #953, opened 2018-12-05
remove sgd gpi entry

kltm:
Remove confusing GPI entry as a workaround for https://github.com/geneontology/neo/issues/36

### DIFF
--- a/metadata/datasets/sgd.yaml
+++ b/metadata/datasets/sgd.yaml
@@ -45,22 +45,22 @@ datasets:
     - NCBITaxon:41870
     - NCBITaxon:4932
     - NCBITaxon:559292
- -
-   id: sgd.gpi
-   label: "sgd gpi file"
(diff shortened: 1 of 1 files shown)


kltm:
@dougli1sqrd If @cmungall is not available, could you confirm this has the desired effect?

dougli1sqrd:
If this passes the schema checker, then this is fine. The pipeline doesn't look at GPI files yet, so this seems all good to me.

kltm:
Hmmm...maybe it's looked at in the NEO assembler. I'll merge and I guess we'll see.

cmungall:
This should have the desired affect, thanks.

https://build.berkeleybop.org/job/build-noctua-entity-ontology/
ran successfully

Connection is too slow to check the file here

We'd expect gene IDs to show up here:

```
$ curl -L -s http://purl.obolibrary.org/obo/go/noctua/neo.owl | grep SGD
```

Issue #956, opened 2018-12-11
GoRules: Yaml should have schema for example data each of a passing and a failing for the Go Rule


dougli1sqrd:
In the GO Rules YAML, let's create specs for optional passing and failing exemplar data for that rule. There could be more than one example of each passing and failing and the data need not be GAF. 

Something like:
```
examples:
    pass:
        - format: rdf|gaf|gpad|etc?
          instance: "blah blah blah"
        - format: ...
    fail:
        - format: rdf|gaf|gpad|etc?
          instance: "blah blah blah"
```

Advantages to this is that it's still format agnostic. We can have an enum of allowed formats in the schema. What formats should be allowed here? Also, not sure if I like `instance` as the name here. Suggestions? The other advantage is that we can have any number of failing and passing examples in different formats. Each example should be expected to "pass" (all failing example entries should fail, all passing entries should pass the rule).

Tagging: @selewis @cmungall @kltm 
```

selewis:
I rather like this idea, or something similar because it makes it easy for curators to contribute their exemplars (in theory, but we do need a solution that enables the most direct communication from them to us as possible)

dougli1sqrd:
Yeah, they could just add it in with a Pull Request. The danger here is having a zillion examples for each rule, and that seems counter-productive. Basically the example should probably illustrate a peculiarity of the rule, exercising a particular condition of the rule.

pgaudet:
@selewis 
I had done actual GAFs for checking the rules (see https://drive.google.com/drive/folders/12H6Kz9BsFEhtllx7CMy9OW29FeiC8GwQ) - can they be translated into this yaml format ? How is this going to work? 

Thanks, Pascale

selewis:
We'll try and load those examples via script since you've already gone to the trouble of creating them. They sparked further discussion, so it was not for naught.

dougli1sqrd:
We have tickets now for adding the example data https://github.com/geneontology/go-site/issues/968, and for creating the software https://github.com/geneontology/go-site/issues/970.

I think this ticket should be closed, as the primary focus here was the schema implementation.

PR #962, opened 2018-12-19
adding a pass, fail, and repair schema to go rule yaml portions

dougli1sqrd:
This is for https://github.com/geneontology/go-site/issues/956. 

The only difference I've done in the schema implementation is to have the key called `repair` rather than transform. If `repair` is agreeable (since that's the language we use for that type of rule already), then this is good. Otherwise I can always change it to previously used `transform`.


kltm:
@dougli1sqrd `repair` would require `input` and `output`.

dougli1sqrd:
oh yes, my mistake.

PR #971, opened 2019-01-07
Update fb.yaml

hattrill:
Changed ftp pick up for gaf, added gp2protein ftp source and change of grant code.


dougli1sqrd:
@hattrill It appears that the stanza you're adding (the gp2protein) is a list item under `datasets`. It just looks like an indentation issue? I think if line 26 is indented to be in line with the first stanza it should pass the travis check.

PR #974, opened 2019-01-10
changing variable names for external exposure

dougli1sqrd:


kltm:
While the Reviewer does not believe that variables should typically expose themselves in public as these wonton variables do--as many regard it as indecent and vulgar--we agree that when necessary exposure occurs, the variables should be names such.

PR #984, opened 2019-01-23
Minor make changes

dougli1sqrd:
This allows the makefile to be run with a parallel mode.


kltm:
Seems to work well so far.

PR #987, opened 2019-01-30
WIP: rule  'NOT' annotation should not have extension

tmushayahama:


dougli1sqrd:
Hi @tmushayahama Can you add a detailed description for this rule beneath the yaml? See gorule 8: https://github.com/geneontology/go-site/blob/master/metadata/rules/gorule-0000008.md as an example.

pgaudet:
@dougli1sqrd This is a different issue. 
However - I think NOT annotations could have extensions, right ? Like 'not apoptotic process in cell type x' 

When did we decide on this rule ? 

Thanks, Pascale

cmungall:
I think @dougli1sqrd was providing rule8 as a template for @tmushayahama to follow

I do not know from where this ticket originated, looks like it was during the hackathon so I assume was driven by people there (I missed the 29th, but was there other days).

I personally think combining extensions with NOT is dangerous. If we forbid NOT then the information is monotonic, i.e. you can omit parts of the annotation and it is still true. E.g. you can ignore c16 (which many do) and you have less information but it's still correct.

Including NOT makes this non-monotonic, and ignoring c16 can flip the interpretation to being wrong.

E.g.

 * I live in a house (that is painted blue) ENTAILS I live in a house
 * I do not live in a house that is painted blue DOES NOT ENTAIL I do not live in a house

dougli1sqrd:
@pgaudet, @cmungall, Indeed, I was supplying rule 8 as an example to Tremayne to update this PR. 

This comes from a couple of rules that Tremayne is adding that he uses in his work in Noctua. This is so that we can record all the implicit rules he's using.

pgaudet:
Hi @cmungall Are you saying that we want to express non-monotonic statements ? 

For example we have annotations where a protein is not found in the nucleus, so it will get an annotation NOT nucleus; but perhaps it would be found in the nucleus of a different cell types. Having that specific information would allow to understand the nature of the conflict in these two annotations.

pgaudet:
Annotations that fail that check are mostly from MGI. I created a list of the annotations for MGI to weigh in on this.

Issue #994, opened 2019-02-11
Update all datasets metadata that use old root domain reference geneontology.org


kltm:
With the switch of the root domain to a new machine, many old mappings no longer work. A critical one, the web version of the SVN repository cannot be mapped over on the root without porting the SVN server or adding two new mappings to bridge.

As a better stopgap solution until we get rid of the dependencies on SVN (see https://github.com/geneontology/go-site/issues/784), we'll make use of the viewvc subdomain, which still works.

kltm:
Mapping like: http://viewvc.geneontology.org/viewvc/GO-SVN/trunk/gene-associations/submission/gene_association.aspgd.gz

Tagging @dougli1sqrd

Issue #784, opened 2018-08-28
Report request - any groups still have SVN as a source for GO data?


vanaukenk:
From USC hackathon - are there any groups whose annotation and/or gene product mapping files (e.g. gpi) are still sourced from the SVN directory?

kltm:
aspgd.yaml:   source: http://geneontology.org/gene-associations/submission/gene_association.aspgd.gz
cgd.yaml:   source: http://geneontology.org/gene-associations/submission/gene_association.cgd.gz
ecocyc.yaml:   source: http://geneontology.org/gene-associations/submission/gene_association.ecocyc.gz
fb.yaml:   source: http://geneontology.org/gene-associations/submission/gene_association.fb.gz
genedb.yaml:   source: http://geneontology.org/gene-associations/submission/gene_association.GeneDB_Pfalciparum.gz
gonuts.yaml:   source: http://geneontology.org/gene-associations/submission/gene_association.gonuts.gz
gramene.yaml:   source: http://geneontology.org/gene-associations/submission/gene_association.gramene_oryza.gz
jcvi.yaml:   source: http://geneontology.org/gene-associations/submission/gene_association.jcvi.gz
pamgo.yaml:   source: http://geneontology.org/gene-associations/submission/gene_association.PAMGO_Atumefaciens.gz
pamgo.yaml:   source: http://geneontology.org/gene-associations/submission/gene_association.PAMGO_Ddadantii.gz
pamgo.yaml:   source: http://geneontology.org/gene-associations/submission/gene_association.PAMGO_Mgrisea.gz
pamgo.yaml:   source: http://geneontology.org/gene-associations/submission/gene_association.PAMGO_Oomycetes.gz
pseudocap.yaml:   source: http://geneontology.org/gene-associations/submission/gene_association.pseudocap.gz
reactome.yaml:   source: http://geneontology.org/gene-associations/submission/gene_association.reactome.gz
rgd.yaml:   source: http://geneontology.org/gene-associations/submission/gene_association.rgd.gz
sgd.yaml:   source: http://geneontology.org/gene-associations/submission/gene_association.sgd.gz
sgn.yaml:   source: http://geneontology.org/gene-associations/submission/gene_association.sgn.gz
tair.yaml:   source: http://geneontology.org/gene-associations/submission/gene_association.tair.gz

tonysawfordebi:
We still submit several external2go (and gp2protein) files to SVN - see also https://github.com/geneontology/go-site/issues/688

[...]

PR #995, opened 2019-02-11
replacing source urls with viewvc urls for svn sources

dougli1sqrd:


kltm:
@dougli1sqrd Noting that there seems to have been more mentioned here: https://github.com/geneontology/go-site/issues/784

dougli1sqrd:
Yep. A few sources have already been fixed since that list was produced, so that's why there are fewer here. Everything else checks out.

kltm:
Note that this PR is for https://github.com/geneontology/go-site/issues/994

Issue #1001, opened 2019-02-16
GO refs do not seem to be properly formatted yamldown/frontmatter


kltm:
Currently, the GO refs in /metadata/gorefs/goref-*.md do not seem to parse properly under any frontmatter/yamldown parser, giving a variably confused mess.

Related to: https://github.com/dougli1sqrd/yamldown/issues/3 and https://github.com/geneontology/go-site/issues/993

kltm:
Whether or not they were proper before, they are at least conforming to a tighter "conforming" subset now.

PR #1002, opened 2019-02-16
possible fixes for the gorefs files for #1001

kltm:
I want to make sure these fixes don't actually mess something else we do with them. Maybe the Makefile?


kltm:
Gunna see what happens with the Makefile~!

Issue #993, opened 2019-02-11
Create new go refs format and generate GO.references


alexsign:
Hi, 
I cannot download following any more.
http://www.geneontology.org/doc/GO.references
It gives 404 error. If it's planned removal, can you please point me to an alternative resource.

kltm:
Duplicate of geneontology/helpdesk#180

kltm:
We used to have a file http://52.27.86.54/doc/GO.references that no longer seems to have an analog. We now produce http://snapshot.geneontology.org/metadata/gorefs/, from the metadata at https://github.com/geneontology/go-site/tree/master/metadata/gorefs . 

We want to:

- [ ] create a new metadata metadata/go-refs.json, compiled from metadata/gorefs/goref-*.yaml
- [ ] convert this file into metadata/GO.references, following to previous format as much as possible
- [ ] add a 302 (temp) forward from /docs/GO.references to /metadata/GO.references

kltm:
http://geneontology.org/doc/GO.references now temporarily working again.

alexsign:
thanks @kltm Can I assume http://snapshot.geneontology.org/metadata/gorefs/ is new permanent location and all gorefs combined file will not be produced in the future?

kltm:
@alexsign 

From https://github.com/geneontology/go-site/issues/993#issuecomment-462557136 , we will be producing a work-a-like for GO.references from a new format that we'll be producing (at a new location). In the interim, we've manually added the old file.
In general, we'd encourage people to use the versioned monthly releases, with the most recent at current.geneontology.org. There are special use cases where people may be interested in snapshot releases as well.

cmungall:
Minor: would be nice to set mime type so it resolves as a text file in a browser rather than defaulting to download

kltm:
Downgraded as temp in place.

kltm:
Temporarily blocked by https://github.com/dougli1sqrd/yamldown/issues/3

kltm:
Final steps are

- [ ] add script to pipeline
- [ ] check that output and map are okay

kltm:
Changes pulled through to snapshot and release. All points hit, now waiting to clear.

PR #1006, opened 2019-02-21
Issue 1003 yamlmatter and consequences

kltm:
As part of the larger program to get GO.references reflecting reality again (see https://github.com/geneontology/go-site/issues/993), we want to have some schema checking on the gorefs. This effort continues with (https://github.com/geneontology/go-site/compare/issue-1003-yamlmatter), but it appears that several of the gorefs do not follow the proposed new schema.

I'd like a review to both okay the schema and the (fairly minimal) changes to the gorefs so that we can proceed with #993 .

### DIFF
--- a/metadata/gorefs.schema.yaml
+++ b/metadata/gorefs.schema.yaml
@@ -1,0 +1,43 @@
+type: map
+mapping:
+  "layout":
+    type: str
+    required: true
+    enum: [goref]
+  "id":
+    type: str
+    required: true
+    pattern: /GO_REF:[0-9]{7}/
+  "authors":
+    type: str
+    required: true
+  "year":
+    type: number
+    required: false
+  "external_accession":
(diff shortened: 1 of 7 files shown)


metadata/gorefs.schema.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - type: str
+        required: false
+        pattern: /[a-zA-Z\_]+:[0-9a-zA-Z\_]+/
+  "alt_id":
+    type: seq
+    required: false
+    sequence:
+      - type: str
+        required: false
+        pattern: /GO_REF:[0-9]{7}/
```
dougli1sqrd: Doe alt_id only get to be a GO_REF ID?

kltm: Apparently there are no counterexamples/

PR #1010, opened 2019-02-26
Update gramene.yaml

suzialeksander:
Updating label field to make http://current.geneontology.org/products/pages/downloads.html uniform, with permission:

Hi Suzi,
Yes it is fine by us. Please proceed. You may have to update the Xref
YAML, as this is a Xref used in other projects as well e.g. Planteome. I
will inform Planteome.
Pankaj

### DIFF
--- a/metadata/datasets/gramene.yaml
+++ b/metadata/datasets/gramene.yaml
@@ -1,5 +1,5 @@
 id: gramene
-label: gramene_oryza
+label: Gramene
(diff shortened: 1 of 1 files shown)


kltm:
@dougli1sqrd Are any of the top levels used for anything in the pipeline as far as identification goes--this shouldn't bork things, right?

dougli1sqrd:
No this should be fine.

PR #1011, opened 2019-03-06
Updated case of entities

pgaudet:


pgaudet:
Do you want to merge?

kltm:
I'm not actually sure about this one. IIRC, the `entity_type`s are controlled labels. As well, isn't `submitter` used in regexps somewhere?

kltm:
I think I'd like a double check from @dougli1sqrd on the `submitter` and to hear from @cmungall on the `entity_type`.

kltm:
Okay, I think we've started talking about this in more detail before, see: https://github.com/geneontology/go-site/issues/1018
Let's continue the conversation over there.

PR #1012, opened 2019-03-06
Changed case dictybase

pgaudet:


kltm:
Assuming @pgaudet knows what's up with dicty.

PR #1015, opened 2019-03-06
create noctua config yaml

dougli1sqrd:


kltm:
@dougli1sqrd Will merging this into master trigger appending the GO-CAM data?

dougli1sqrd:
No. We would still have to release ontobio, and update the versions on go-site. But noctua will be picked up by the pipeline. I'm testing how the pipeline will see noctua as its own group now.

dougli1sqrd:
Adding this won't change anything. This will become a target by the pipeline, but as all sources are gpad, each dataset is skipped. The pipeline currently only accepts gaf primary sources.

kltm:
Fair enough, but should all start as `inactive`, as that will be the initial default as it will be opt-in?

dougli1sqrd:
we currently ignore the `status` field. For documentation purposes we could switch to inactive?

kltm:
@dougli1sqrd Okay to merge if you are okay with my changes.

dougli1sqrd:
Looks good

PR #1017, opened 2019-03-07
allowed reactome to be uncompressed with their new url

dougli1sqrd:


dougli1sqrd:
We don't have to merge this, but this is an example of how this could work with uncompressed gaf sources. If we like it though, might as well merge.

vanaukenk:
@dougli1sqrd 
What is the current status of this pull request?  
Do we need to ask Reactome to gzip their file, or can we go ahead and update the Reactome yaml with the new source URL with an unzipped file?
Thx.

kltm:
@vanaukenk Once @dougli1sqrd changes propagate, this can be merged and we can start using the new URL as source (I'll leave it to @dougli1sqrd to decide the timing as he will be making the final changes).

That said, while we will be able to do it and that should be ready in days, I would still make the not-quite-strong-but-a-little-more-than-passing suggestion that people use gzip compression. I'll leave it to you to make that suggestion or not; otherwise this will go in naturally sooner rather than later.

PR #1019, opened 2019-03-11
Add new source for gaf

glwinsor:


kltm:
@glwinsor The location that you gave is not actually a downloadable file, but a web page.
Perhaps you meant something like: https://github.com/brinkmanlab/pseudomonas-genome-database/raw/master/pseudocap-annotation-files/gene_association.pseudocap.gz ?

PR #1022, opened 2019-03-12
Update reactome.yaml

vanaukenk:
Updating source file location
@dougli1sqrd - please double-check and make sure everything is okay.  Thx.

### DIFF
--- a/metadata/datasets/reactome.yaml
+++ b/metadata/datasets/reactome.yaml
@@ -16,7 +16,7 @@ datasets:
    dataset: reactome
    submitter: reactome
    compression: gzip
-   source: http://viewvc.geneontology.org/viewvc/GO-SVN/trunk/gene-associations/submission/gene_association.reactome.gz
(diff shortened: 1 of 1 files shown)


dougli1sqrd:
This 404s actually https://www.reactome.org/download/current/gene_association.reactome.gz

kltm:
I would note that currently nothing like that seems to exist in https://www.reactome.org/download/current/

ukemi:
It is there on their downloads page, but not as a zipped file.

vanaukenk:
Ah - I think I misunderstood Joel's email.
This file *will* be available with the next Reactome release, 68.
I'll confirm with Joel about when this will be available and then we can merge the pull request.

vanaukenk:
Next Reactome release date: ~March 20th

vanaukenk:
The new Reactome release is available, so the annotation file is now on their site.

I'm going to go ahead and merge the pull request.

PR #1025, opened 2019-03-13
Added Roslin Institute

pgaudet:


metadata/groups.yaml
```diff
@@ -394,3 +394,8 @@
   label: JCVI
   id: https://www.jcvi.org
   shorthand: JCVI
+-   
+  label: Roslin Institute
+  id: https://www.ed.ac.uk/roslin
+  shorthand: RI
```
dougli1sqrd: So, just for reference, `assigned_by` checks the `shorthand` field for a match. In the offending GAF that spawned this change, the `assigned_by` is ` Roslin_Institute`. This will continue to show up as an error as written here. Are we saying that `RI` is the canonical form, and that ` Roslin_Institute` should be changed in the GAF?

kltm:
As we have Roslin_Institute in the GAFs, let's just got for that for now--if we need a change to something more "correct" we can do that later.

pgaudet:
Hi @dougli1sqrd 
Can we update the documentation to state which field should be used ? Or how are people supposed to know ?

PR #1026, opened 2019-03-13
Removed Roslin institue

pgaudet:


kltm:
This information would seem to be fine to hold for now; might even be useful for linkouts.

pgaudet:
OK - I thought @cmungall said that only groups who mint IDs should be there. 

Can you give some documentation of the scope of this file ?

dougli1sqrd:
This is what the metadata readme says: https://github.com/geneontology/go-site/blob/master/metadata/README.md#groupsyaml

pgaudet:
"Groups encompasses organizations, projects, working groups, content meetings, grants, etc. " ... ?

cmungall:
@dougli1sqrd - you are pointing to the documentation for the groups.yaml file, whereas this PR pertains to the db-xrefs yaml file.

It should be simple

 * db-xrefs.yaml should be for anything that is col1 of a GAF, or the prefix of a CURIE/Identifier
 * groups.yaml for anythig that cann be the value of a contributor or assigned_by

Now, in the past we may have conflated these, and it may be the case that P2GO makes use of db-xrefs.yaml to check the assigned_by field. @alexsign would you have an issue with us merging this PR?

dougli1sqrd:
@cmungall, Oh sorry, I was confused by @pgaudet's comment, asking you about a previous comment you made about *groups* - so I linked to the documentation about the groups.yaml.

> It should be simple
> 
>     db-xrefs.yaml should be for anything that is col1 of a GAF, or the prefix of a CURIE/Identifier
>     groups.yaml for anythig that cann be the value of a contributor or assigned_by

That is what we're doing. 

I think @kltm closed this PR because in the email chain where we CC'd Alex Ignatchenko about this, he stated:

> I don't see any issue related to assigned_by validation from our side. However, we do use db-xrefs.json a lot for things like synonyms, id_syntax and url.  I might not understand full story behind this move, but would it be better to leave Roslin_Institute in db-xrefs file for purposes above and add record to groups.yaml for consistence with the other groups/databases.

I otherwise don't know about documentation for dbxrefs, sorry for the red herring.

cmungall:
@alexsign there are no IDs with the prefix Roslin_Institute, my inclination is to delete from db-xrefs and keep the record (plus syns, URLs) in one place. It sounds like this won't affect you?

alexsign:
@cmungall yes, it shouldn't affect us.

pgaudet:
@kltm 's comment when he closed the ticket was 
> This information would seem to be fine to hold for now; might even be useful for linkouts.

I think it's confusing to have stuff in that file that should not be there according to our documentation; if you're OK to remove Roslin Institute form that file please merge my PR. 

Thanks, Pascale

dougli1sqrd:
Are we okay to merge this then?

pgaudet:
It would be OK for me !

pgaudet:
Removed via another PR

PR #1027, opened 2019-03-13
adding schema for a list of tags in rules, only allowing gzip if any for datasets

dougli1sqrd:
Precursor for silencing gorule 26. Also adds the change so only gzip or empty is allowed for the `compression` field in datasets.


kltm:
Noting there is also a `wget` change.

kltm:
It looks like `tags` will be the way that we sort. Is there an enummed set for this yet?

Issue #741, opened 2018-08-02
GORULE:0000023 'Materialize annotations for inter-branch links in the GO' is not a rule


pgaudet:
GORULE:0000023  Materialize annotations for inter-branch links in the GO
This is very interesting - however it should go in other documentation, possibly the pipeline documentation ? where we document our products. 
People need to know that this data in being generated, how, and how to consume it. 

2 actions: 
1. Move this to documentation (Pascale) 
2. Deprecate this as a rule (Eric)

pgaudet:
@cmungall Is the text that is there OK to start the documentation ? Or please provide some text.

cmungall:
This is a rule. We can add a field that distinguishes constraint rules from generative rules.

PR #1032, opened 2019-03-19
This changes status to deprecated and removes the listed issue

dougli1sqrd:
From https://github.com/geneontology/go-site/issues/741


pgaudet:
Hi @cmungall 
I had proposed that the predictions made by inter-ontology be moved elsewhere - but where can they move ? As long as they don't have a new place to go we cannot merge this. 

Thanks, Pascale

Issue #790, opened 2018-08-29
Implement gorule-0000043: Allowed References for each ECO


pgaudet:

pgaudet:
@dougli1sqrd 
Can I change the format of the GO_refs to add which evidence codes are allowed ? Would that be a way to provide the information? 

See for ex: 
https://github.com/geneontology/go-site/blob/master/metadata/gorefs/goref-0000037.md

--- 
authors: UniProt-GOA
external_accession: 
  - "SGD_REF:S000148671 "
id: GO_REF:0000039
year: 2011
layout: goref
**+ ADD evidence_codes: ECO:000314, ECO:0002221, etc**
---

Would that work ? 

Thanks, Pascale

dougli1sqrd:
I like this idea. I'll make a PR for the reference schema.

dougli1sqrd:
https://github.com/geneontology/go-site/pull/1033

dougli1sqrd:
@pgaudet your new goref field is now available!

pgaudet:
thanks !

dougli1sqrd:
This is complete. The current effort is here: https://github.com/geneontology/go-site/issues/1177

PR #1033, opened 2019-03-19
adding allowed evidence to gorefs schema

dougli1sqrd:
This adds machine readable yaml for the go references for allowed ECO IDs used with a particular GO_REF. https://github.com/geneontology/go-site/issues/790#issuecomment-425160398


kltm:
Looking at reference, looks good to me. Merge if pass.

PR #1048, opened 2019-03-29
Change base URL for ZFIN

lpalbou:
@cmungall could you check this ? It is to answer an AGR demand to correct and simplify ZFIN URLs in the Ribbon (which uses amigo2-instance-data which uses this file): https://agr-jira.atlassian.net/browse/AGR-1350


cmungall:
I believe the example_id field should be the local_id (although I see now we are not consistent here)

PR #1052, opened 2019-04-02
Update tair.yaml

tberardini:
File no longer committed to svn but automated to be deposited at this URL on a quarterly basis.


tberardini:
@kltm @vanaukenk We've changed our file location as requested. Please check that I updated this yaml file correctly.

kltm:
Looks good to me.

PR #1058, opened 2019-04-17
align users.yaml and groups.yaml

kltm:
attempted fix for https://github.com/geneontology/noctua/issues/610 using groups.yaml as the invariant


kltm:
Minor changes, I'm going to self-review.

Issue #1066, opened 2019-04-26
Filter out CGD and AspDB from goa_uniprot_all.gaf


pgaudet:
Hello, 

We are loading CGD and AspDB twice: once from their own sources, and once from UniProt-all. We need to filter them out from UniProt-all. 

For example: http://amigo.geneontology.org/amigo/gene_product/CGD:CAL0000173998
and 
http://amigo.geneontology.org/amigo/gene_product/UniProtKB:Q5ABU7

Thanks, Pascale

pgaudet:
(this task is coming from #997)

kltm:
Source is goa_uniprot_all-src.gaf.gz

kltm:
Will need to add a GO Rule accordingly.

kltm:
Let's take an example line coming in:
`UniProtKB       Q5ABU7  MDR1            GO:0005887      PMID:20041202   IDA             C       Multidrug resistance protein 1  MDR1|BEN|BMR1|CAALFM_C603170CA|CaO19.13047|CaO19.5604   protein taxon:237561    20140625        CGD             `
Okay:
```
sjcarbon@moiraine:/tmp$:) grep "[[:space:]]CGD[[:space:]]" goa_uniprot_all_noiea.gaf |  cut -f 13 | sort | uniq -c | sort -n
     35 taxon:573826
   4383 taxon:284593
   5450 taxon:578454
  14522 taxon:237561
```
These strains should be added to cgd.yaml
Similarly:
```
sjcarbon@moiraine:/tmp$:) grep "[[:space:]]AspGD[[:space:]]" goa_uniprot_all_noiea.gaf |  cut -f 13 | sort | uniq -c | sort -n
  12718 taxon:330879
  13851 taxon:510516
  14628 taxon:227321
  23217 taxon:425011
```
These strains should be added to aspgd.yaml

pgaudet:
@marekskrzypek 
Can you export the above-mentioned taxa ?

kltm:
The greps in https://github.com/geneontology/go-site/issues/1066#issuecomment-487218802 now return no results.

PR #1067, opened 2019-04-26
attempt to fix filters for #1066

kltm:


kltm:
@cmungall We need this checked for #1066

kltm:
Moving ahead.

PR #1073, opened 2019-05-02
Update users.yaml

suzialeksander:
adding new SGD curator


kltm:
Assuming @suzialeksander knows who their curators are.

PR #1074, opened 2019-05-07
goa all awk script now outputs a basic report on all skipped lines

dougli1sqrd:


kltm:
Assuming this is as discussed yesterday.

Issue #1072, opened 2019-05-02
Make use of new standalone ontology products in pipeline


balhoff:
The "pre-merged, standalone ontologies" ticket is done now in go-ontology: https://github.com/geneontology/go-ontology/issues/16876

I'm not sure what needs to change here, but this is just a request to point directly to the newly built ontology files for go-plus or go-gaf for use in the later pipeline steps (instead of referencing them via PURL). These now have all imports merged in, so no use of a catalog is needed.

dougli1sqrd:
https://github.com/geneontology/go-site/pull/1076
This is clearing now, let's wait for the next run and we can close it tomorrow.

pgaudet:
Can this be closed ? Does this belong to a project (for reporting purposes)

kltm:
@pgaudet This cannot be closed.

PR #1075, opened 2019-05-07
pipeline now just downloads the big ontology instead of using owltools to grab imports

dougli1sqrd:
For https://github.com/geneontology/go-site/issues/1072


kltm:
As the target seems the same, assuming no real difference here (besides not dealing with imports).

PR #1076, opened 2019-05-07
switching to snapshot url

dougli1sqrd:


pipeline/Makefile
```diff
@@ -44,7 +44,7 @@ target/go-ontology.json:
 
 GAF_OWL = target/go-gaf.owl
 $(GAF_OWL):
-	wget --no-check-certificate $(OBO)/go/extensions/go-gaf.owl -O $@.tmp && mv $@.tmp $@ && touch $@
+	wget --no-check-certificate $(OBO)/go/snapshot/extensions/go-gaf.owl -O $@.tmp && mv $@.tmp $@ && touch $@
```
balhoff: @dougli1sqrd @kltm just curious will this point to the ontology file just built within the same run? Or is it the previous successful snapshot?

kltm: The latter apparently. For doing the "single pass downloads" this would have to reference an internal URL. The is no worse than what we had before.

balhoff: I guess so but it seems like this will continue some of the mysterious behavior that requires two runs to get right.

PR #1080, opened 2019-05-08
fixing broken zip product in goa_uniprot_all

dougli1sqrd:
For finishing off https://github.com/geneontology/pipeline/issues/93

This correctly `mv`s the .tmp file into the real named file, that the gzip command is expecting.

This was not caught because there were stale products already existing with the right name from older local builds on my machine.


kltm:
Assuming things are where I think they are, looks good.

Issue #1082, opened 2019-05-08
Refine gorule-0000033 to be a warning and the wording relaxed and ontobio impl


dougli1sqrd:
We should refine https://github.com/geneontology/go-site/blob/master/metadata/rules/gorule-0000033.md so that fail_mode=soft, because there is too high a chance of removing valid annotations (e.g. if a group provides a MOD ID but no PMID)

gorule 33 currently is worded very aggressively. The wording implies that we can understand which IDs are synonyms. 

We should change this to assume that different IDs *are* synonyms, and then `repair` the reference that is the DB ID by removing it, leaving only the GO_REF or PMID, etc (full list is in the current go rule).

Closing this ticket comes along with an ontobio implementation.
(See also https://github.com/geneontology/go-site/issues/1063#issuecomment-490651143)

pgaudet:
We need to fist get the MODs to remove their publication IDs from the files they provide. 

We should focus the work on rules that are 'approved' (or incorrectly implemented). 

Thanks, Pascale

dougli1sqrd:
The implementation of this rule will remove the Group specific ID. This is only coming up because of https://github.com/geneontology/go-site/issues/1063#issuecomment-490651143.

pgaudet:
I changed the status to approved. 

We should notify contributing groups. Who can do that ?

dougli1sqrd:
https://github.com/geneontology/go-site/pull/1083

dougli1sqrd:
@pgaudet I'm not sure how that would work, what are the normal channels to do this?

pgaudet:
Annotation calls, go-consortium email list, and go-announcements repo.

dougli1sqrd:
https://github.com/geneontology/go-site/pull/1197

pgaudet:
* The current proposal is to filter internal MOD reference ID for GO_REFs - ie if there are both a internal MOD reference ID  and a GO_REF, we will filter out the MOD ref ID. 
* For PMIDs also having a internal MOD reference ID, both references will be kept for now.  

@vanaukenk does that seem to reflect the decision from the Berkeley GOC meeting ? 

Thanks, Pascale

pgaudet:
After discussion with @ukemi and @vanaukenk 
We will deprecate this rule. Groups want to link to their internal pages for all references. 

@sabrinatoro I think ZFIN removed those already; if you want to you can add them back. Sorry about that ! 

Thanks, Pascale

[...]

PR #1083, opened 2019-05-09
update wording in gorule-33

dougli1sqrd:
See https://github.com/geneontology/go-site/issues/1082

### DIFF
--- a/metadata/rules/gorule-0000033.md
+++ b/metadata/rules/gorule-0000033.md
@@ -2,7 +2,7 @@
 layout: rule
 id: GORULE:0000033
 title: Public Reference IDs (PMID, PMC, doi, or GO_REF) should be preferred over group specific Reference IDs
-type: repair
+type: report
 fail_mode: soft
 status: implemented
 contact: "go-quality@mailman.stanford.edu"
--- a/metadata/rules/gorule-0000033.md
+++ b/metadata/rules/gorule-0000033.md
@@ -1,13 +1,13 @@
 ---
 layout: rule
 id: GORULE:0000033
-title: Group specific Reference IDs (column 6) will be replaced by corresponding GO_REF (or other public ID) or filtered.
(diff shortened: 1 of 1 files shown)


dougli1sqrd:
also, @cmungall can we change the rule to `approved` or `implemented`? Right now it's still proposed.

pgaudet:
Sorry @dougli1sqrd I accidentally changed the status to 'approved' directly into master. (I should go to bed ;)

dougli1sqrd:
This is implemented here: https://github.com/biolink/ontobio/pull/327

kltm:
I would think that "should be preferred" is a little weaker than the "will be filtered" rule that it is...

dougli1sqrd:
Well since this is mostly a report, and a repair only if we can, I worded it more gently. Since we won't filter out "bad" references.

vanaukenk:
This rule is part of the GPAD 2.0 specifications, but I didn't know that we were starting to implement it now.
We should probably discuss on an annotation call to give people more notice.

dougli1sqrd:
Yeah, discussing it on the next call is fair. We can at least merge these changes to this rule if we agree with the wording, and not update the pipeline to the latest ontobio so the changes don't actually get activated until we're ready? @kltm does that or something similar sound like an okay way to go?

kltm:
@dougli1sqrd I would actually defer merging until we were ready (as that may end up blocking other things in the interim).
Make sure the PR is tied to a ticket and we can revisit.

kltm:
@pgaudet @vanaukenk If data is being changed for our downstream bulk consumers (as opposed to "direct" AmiGO users) and it is not a critical "damage" fix, perhaps they could be introduced first as a warning/report and then graduate to mutation/filtering once there was a cool-off period (1m? 3m?)?

dougli1sqrd:
I updated the PR to just report, and changed the ontobio PR implementation here: https://github.com/biolink/ontobio/pull/328 to just WARN also

PR #1090, opened 2019-05-14
removing taxon:4081 from sgn metadata

dougli1sqrd:
Here's the change to remove NCBITaxon:4081 from sgn to stop filtering taxon:4081 from goa_uniprot_all.


kltm:
In reference to https://github.com/geneontology/pipeline/issues/92
Allow tomato annotations from other locations--SGN no longer canonical resource.

PR #1104, opened 2019-06-13
Updated RefGenome URL

pgaudet:


kltm:
@pgaudet Changing `id`s in this file is equivalent to changing GO ids in the ontology and either due diligence that it has not already been used or a migration plan needs to be done before changing ids.

pgaudet:
So I should send a notification on GO-friends and on the go-announcemets repo ? 

I just hope nobody goes to the URL provided -  https://cran.r-project.org/web/packages/refGenome/index.html

This is completeley unrelated to the RefGenome/PAINT project 
refGenome: Gene and Splice Site Annotation Using Annotation Data from 'Ensembl' and 'UCSC' Genome Browsers

Contains functionalities for importing and managing downloaded genome annotation data from 'Ensembl' genome browser (European Bioinformatics Institute, <https://www.ensembl.org>) and from 'UCSC' genome browser (University of California, Santa Cruz, <https://genome.ucsc.edu/>) and annotation routines for genomic positions and splice site positions.

kltm:
This is more along the lines of one would need to scan through all of the GO-CAM models to see if somebody has used this as their group.
Again, this is *not* an "informational" URL, but an (in most cases) immutable id; in fact, it may not resolve at all in some cases. If we want to add a separate `homepage` URL for groups, that would be fine and could be something more useful for look ups.
Tagging @cmungall

cmungall:
Wow, we should never have had a cran URI in the first place

Yes in general we should not change but I think we need a one time change to make these more sensible, so long as not used in noctua annotation should be fine

We should also have a homepage field for a mutable informative URL

PR #1128, opened 2019-06-25
Update group-contacts.csv

bmeldal:
Complex Portal was missing, I was only down for IntAct.


pgaudet:
This is in the file (maybe via another PR?)

PR #1131, opened 2019-06-28
Updated EC links

pgaudet:
@kltm Are we allowed to put more than one URL example? It's nice to know how to construct partial links. (We have many links missing the 'dash' in the ontology- but that's another issue).

### DIFF
--- a/metadata/db-xrefs.yaml
+++ b/metadata/db-xrefs.yaml
@@ -641,7 +641,6 @@
       url_syntax: http://www.expasy.org/EC/[example_id]
       example_id: EC:1.4.3.6
       example_url: http://www.expasy.org/enzyme/1.4.3.6
-      example_url: http://www.expasy.org/enzyme/1.4.3.-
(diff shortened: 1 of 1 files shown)


kltm:
@pgaudet 
https://www.expasy.org/EC/1.4.3.6
https://www.expasy.org/EC/1.4.3.-
Are not legal URLs. What is your intention with the PR? The syntax and examples are now different.

kltm:
examples are singular in the schema:
https://github.com/geneontology/go-site/blob/b1694ab5de6edfc0b3d4fada97cc5f9b366f508c/metadata/db-xrefs.schema.yaml#L57

pgaudet:
Thanks for the feedback @kltm

kltm:
@pgaudet Could you please make it so the example URL and the URL giving the syntax are of the same format? As it stands now, the testing URL cannot be generated from the syntax URL.

PR #1134, opened 2019-07-01
Issue 1133 downloader script

dougli1sqrd:


kltm:
The proof will be in the pudding with how this interacts with the rest of the pipeline.

PR #1138, opened 2019-07-03
script is now more careful if it encounters a strange file in organiz…

dougli1sqrd:
…ing.


kltm:
More checks are good.

Issue #842, opened 2018-10-04
Add GO release stats to pipeline


pgaudet:
Hello, 

It'd be really useful to have release stats at each GO release (snapshots too, so we could look for anomalies before official releases).

We use to have a lot of queries from GOOSE (I guess), here: 
http://geneontology.org/page/go-mysql-database-schema-views

pgaudet:
Full list is: 
go_annotation_reports

    annotated_publication_total
    annotated_publication_total_by_evidence_code_non_additive
    annotated_total_gps_by_evidence_code_non_additive
    association_contradiction
    association_contradiction_count_by_ontology
    association_contradiction_direct
    association_contradiction_direct_count_by_ontology
    association_count_by_association_qualifier
    association_count_by_association_qualifier_and_dbname
    association_total_by_evidence_code
    association_total_by_evidence_code_and_species
    avg_total_annotations_per_gp_by_db
    avg_total_nonroot_annotations_per_gp_by_db
    avg_total_nonroot_pubs_per_gp_by_db
    avg_total_nonroot_terms_per_gp_by_db
    avg_total_nonroot_transitive_terms_per_gp
    avg_total_nonroot_transitive_terms_per_gp_by_db
    avg_total_pubs_per_gp_by_db
    avg_total_terms_per_gp_by_db
    avg_total_transitive_terms_per_gp
    avg_total_transitive_terms_per_gp_by_db
    evidence_dbxref_summary
    evidence_pub_dbxref_summary
    gene_product_dbxref_summary
    iea_annotated_total_gps
    iea_or_iss_annotated_total_gps
    non_iea_annotated_total_gps
    non_iea_annotated_total_gps_by_dbname
    non_iea_or_iss_annotated_total_gps
    ont_association_count_by_association_qualifier
    seq_dbxref_summary
    term_association_count_by_association_qualifier
    term_association_count_by_fraction_type
    term_association_count_by_fraction_type_and_evidence
    term_dbxref_summary
    total_annotated_entities_by_dbname_and_type
    total_annotations_per_gp
    total_gps_by_dbname
    total_nonroot_annotations_per_gp
    total_nonroot_pubs_per_gp
    total_nonroot_terms_per_gp
    total_nonroot_transitive_terms_per_gp
    total_nonroot_transitive_terms_per_gp_pair
    total_pubs
    total_pubs_per_gp
    total_terms_per_gp
    total_transitive_terms_per_gp
    total_transitive_terms_per_gp_ont

go_associations

[...]

PR #1145, opened 2019-07-17
go-stats : compute stats on go annotations from a GOLr instance

lpalbou:
Note the full code as well as the PyPI setup to publish the package is also here: https://github.com/geneontology/go-stats/tree/master/libraries/go-stats

### DIFF
--- a/scripts/go_stats.py
+++ b/scripts/go_stats.py
@@ -1,0 +1,480 @@
+# GO Update Statistics
+
+import requests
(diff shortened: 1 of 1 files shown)


kltm:
Noting that the code always references the current `release`, as discussed.

kltm:
Need to tie to ticket
https://github.com/geneontology/go-site/issues/842

PR #1150, opened 2019-07-29
removing gaferencer target and dependencies in makefile

dougli1sqrd:


kltm:
@dougli1sqrd Noting that this still uses PYTHON_CMD. That's the new one that we want to revert, right?

kltm:
For https://github.com/geneontology/pipeline/issues/116

Issue #1153, opened 2019-08-13
Pipeline Makefile has apparent race condition with go-ontology.json target


dougli1sqrd:
The Makefile in pipeline has a target for the json ontology:
```
target/go-ontology.json:
	wget --no-check-certificate $(ONTOLOGY) -O $@.tmp && mv $@.tmp $@ && touch $@
```

Later on, we use the now mustly defunct rdfgen.py of ontobio on the _noiea goa_uniprot_all product to generate Turtle for making "degenerate" GO-CAM models of the GAF. This rule is:
```
RDF_GEN = $(PY_BIN)/rdfgen.py
target/groups/goa/goa_uniprot_all_noiea_cam.ttl: target/groups/goa/goa_uniprot_all_noiea.gaf
	python3 $(RDF_GEN) convert -a gaf -r target/go-ontology.json -o target/groups/goa/goa_uniprot_all_noiea_cam.ttl target/groups/goa/goa_uniprot_all_noiea.gaf
	gzip -cf target/groups/goa/goa_uniprot_all_noiea_cam.ttl > target/groups/goa/goa_uniprot_all_noiea_cam.ttl.gz
```

However the usage of `-r target/go-ontology.json` is depending on the file but the rule doesn't specify it as a dependency. This seems to be a race condition, as we run this with multiple processes (`make -j 10`). If rdfgen is run before go-ontology.json is downloaded we can have an error (as seen before). 

We should add `target/go-ontology.json` as a dependency. 

Also, please add any discussion as to why we have never run into this before after having had this formed this way for 8 months at this point. Is there any way something like this could happen w/out a race like condition? The context being we have a string of pipeline Jenkins failures and it's very coincidental that it's happening now. But perhaps it is just that 1 in 300 thing!

Tagging @kltm

dougli1sqrd:
This has run in snapshot successfully a handful of times. This seems to be working.

PR #1154, opened 2019-08-13
add target/go-ontology.json as a dependency

dougli1sqrd:
For https://github.com/geneontology/go-site/issues/1153

I guess this "solves" it. I still feel spooked though that it's never been a problem before until now.


kltm:
Best guess is that this should not cause any knock-on problems, and could possibly insulate us from other issues (race or new internal ordering).

PR #1155, opened 2019-08-14
Update go logo URLs

lpalbou:
This template was using old URLs no longer available since the Drupal switch to GH pages


kltm:
Looks great to me.

Issue #1156, opened 2019-08-15
Dual Taxon check gorule-0000015 should split out general GP/taxon verification


dougli1sqrd:
The current GORULE:0000015 text basically outlines 2 checks: 1) A taxon should be aligned with the GP in the annotation, and 2) Dual taxons are only allowed with descendants of GO:0044419 and GO:0044215. 

It seems only point 2 is the actual *dual species check*. Point 1 should be spun out into its own rule  as it should apply to both single taxon fields *and* dual taxons.

dougli1sqrd:
https://github.com/geneontology/go-site/pull/1157

kltm:
We all talked (@pgaudet @kltm @dougli1sqrd ), we'll defer on the proposed actionable for checking taxon against source until later.

PR #1157, opened 2019-08-15
add a new rule, 55, describing the basic taxon relationship with the …

dougli1sqrd:
…GP, edited 15

Basically the first part of the gorule 15 defines that the first taxon should match the species of the GP in the annotation. This is general to both dual taxons and single. So this could be it's own rule. I have edited rule 15 to reflect this change as well. Rule 15 would be *just* checking that dual taxon IDs are present only with the given GO Rules and their descendants.


kltm:
Apoloiges--on a call and we're going for it.

kltm:
https://github.com/geneontology/go-site/issues/1156

PR #1167, opened 2019-09-06
Update dbxrefs.yaml for WB identifiers, URLs

vanaukenk:
@kltm 
I've updated the entries for the WB URLs and IDs.
Three things to check with you, though:
1) I don't have a type_id for RNAi.  These identifiers refer to curated RNAi experiments but searching for 'RNAi experiment' on Ontobee uncovered this:  http://bioportal.bioontology.org/ontologies/EDAM?p=classes&conceptid=topic_3523
Would that possibly work?
2) The example IDs are in the format WB:blah but the actual example URL strips the WB which means that in the URL syntax we're not really using the full example_id.  Is that okay?
3) I did update the id_syntax wrt the protein identifiers because we no longer use the WP: prefix in WB so these would definitely not be correct.  Please let me know, though, if I should revert these back.
Thx.

### DIFF
--- a/metadata/db-xrefs.yaml
+++ b/metadata/db-xrefs.yaml
@@ -3291,22 +3291,34 @@
   entity_types:
     - type_name: gene
       type_id: SO:0000704
-      id_syntax: (WP:CE[0-9]{5})|(WB(Gene|Var|RNAi|Transgene)[0-9]{8})
-      url_syntax: http://www.wormbase.org/db/gene/gene?name=[example_id]
+      id_syntax: (CE[0-9]{5})|(WB(Gene|Var|RNAi|Transgene)[0-9]{8})
+      url_syntax: http://www.wormbase.org/get?name=[example_id]
       example_id: WB:WBGene00003001
-      example_url: http://www.wormbase.org/db/get?class=Gene;name=WBGene00003001
+      example_url: http://www.wormbase.org/get?name=WBGene00003001
     - type_name: variation
       type_id: VariO:0001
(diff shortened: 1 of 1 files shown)


vanaukenk:
@kltm - instant Travis check failure, but I'm not sure how to interpret the error log.

metadata/db-xrefs.yaml
```diff
@@ -3291,22 +3291,34 @@
...
-      id_syntax: (WP:CE[0-9]{5})|(WB(Gene|Var|RNAi|Transgene)[0-9]{8})
-      url_syntax: http://www.wormbase.org/db/gene/gene?name=[example_id]
-      example_id: WB:WBGene00003001
-      example_url: http://www.wormbase.org/db/get?class=Gene;name=WBGene00003001
+      id_syntax: (CE[0-9]{5})|(WB(Gene|Var|RNAi|Transgene)[0-9]{8})
+      url_syntax: http://www.wormbase.org/get?name=[example_id]
+      example_id: WB:WBVar00089843
+      example_url: http://www.wormbase.org/get?name=WBVar00089843
+    - type_name: RNAi
+      type_id:
```
kltm: @vanaukenk I suspect that this being empty is not an option

kltm: https://travis-ci.org/geneontology/go-site/builds/581772518#L1068

vanaukenk:
Checking with @alexsign to make sure proposed changes to WB entries will not cause problems for the EBI.

PR #1169, opened 2019-09-09
add config change and downloader change

dougli1sqrd:
This is for https://github.com/geneontology/pipeline/issues/122. An incoming change in ontobio is also needed so that ontobio respects the `exclude` key in the config. 

This adds `exclude` in datasets yaml as a default, so that these defaults are no longer secret in the Makefile. 

Additionally, the downloader now respects the exclude key.

### DIFF
--- a/metadata/datasets.schema.yaml
+++ b/metadata/datasets.schema.yaml
@@ -109,3 +109,6 @@ mapping:
             required: false
             sequence:
               - type: str
+          "exclude":
+            type: bool
+            required: false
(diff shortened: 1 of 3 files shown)


kltm:
@dougli1sqrd This looks about right, but let's hold on it for the moment until we have a `release` out.

dougli1sqrd:
@kltm We also need this to be merged for the Jenkinsfile downloader branch to work for snapshot.

kltm:
I should read better.

Issue #1165, opened 2019-09-04
Add the Rule title in the large Reports page and possibly in dataset report.


dougli1sqrd:
From https://github.com/geneontology/go-site/issues/866

The column with the rule ID should perhaps also have the title. Or somewhere more accessible than the mouseover - but we will keep the mouse over title text.

PR #1173, opened 2019-09-13
titles are in gorule report page

dougli1sqrd:
This is for https://github.com/geneontology/go-site/issues/1165


dougli1sqrd:
Pascale approves: https://github.com/geneontology/go-site/issues/1165#issuecomment-531115474

PR #1176, opened 2019-09-18
Update gorule 33 description to say that it only reports

dougli1sqrd:
The current version of the rule says that we repair and remove bad refs, but we don't, and the rule type is a report only, so the description should not claim that we repair or modify the annotation in any way.


kltm:
I'm leaving this to @pgaudet .

Issue #1168, opened 2019-09-09
Update pipeline for new GOA tRNA resources to no IEA


kltm:
Recently, it was discovered that GOA tRNA IEAs were skewing the expected GO release statistics. This will need to be remedied before a release can move forward.

Tagging @vanaukenk @pgaudet

kltm:
Change https://github.com/geneontology/go-site/blob/master/metadata/datasets/goa.yaml#L836 to non IEA variant?

pgaudet:
https://github.com/geneontology/go-site/pull/1170

pgaudet:
GOA is now providing rna_noiea files. I have upated the yaml to reflect the file download location.

pgaudet:
Yes. Cow had the largest change, but to be consistent we're removing all
IEA annotations to RNAs.

Thanks, Pascale

pgaudet:
Hi @kltm 

I am not sure I understand the difference between the two options, but the latter sounds more correct, "change the naming and drop the old products."

Thanks, Pascale

kltm:
Talking to @pgaudet , we'll switch to the new identities, with an associated announcement for the next release.

pgaudet:
Announcement is here: https://github.com/geneontology/go-announcements/issues/77

dougli1sqrd:
So we're renaming the dataset IDs in goa.yaml to be the original IDs, but point to the noiea version of the source file. We discovered additional issues in the pipeline related to the renaming of these datasets. So to make the metadata consistent, just changing the source URL and not the IDs was decided as an easier solution. PR here: https://github.com/geneontology/go-site/pull/1184

pgaudet:
Out of date. We had changed the files were are loading from the complete file to the 'no-IEA'. 
However, more recently we have switched to the 'GOA-plus files that include all annotations for gene-centric entities, and now we do have some IEAs to RNAs (although not as many as when we were loading the complete file),

Issue #804, opened 2018-09-11
Update "url" field in datasets to refer to appropriate download locations


kltm:
Currently, the field "url" contains dummy values. We'd like to switch these to the correct `current` pipeline output values and remove the field from elements that will not have downloads made available.

- [x] update schema to optional
- [x] update values
- [x] check downloads page generator

For reference:
https://github.com/geneontology/go-site/issues/753
https://github.com/geneontology/go-site/pull/803

pgaudet:
Isn't this a showstopper... ? displaying dummy values ?

kltm:
This is not a showstopper, it is a small update to a field that only has a single use and is not in production yet.

kltm:
@pgaudet Can you think of any reason that the GOA GAFs have the old SVN file location as their canonical pickup location?

pgaudet:
No 

@alexsign can you please help ? I thought I got this information from you. 

Thanks, Pascale

kltm:
@pgaudet These are likely just the traditional values that we had for the pickups. I'm wondering if there is any particular reason that we didn't migrate them, or if we just somehow overlooked them. I'm almost certain it's the latter, especially if nothing comes to mind for you.

pgaudet:
I don't see what you mean - can you give me an example of the old URLs?

kltm:
url: http://geneontology.org/gene-associations/goa_human_rna.gaf.gz
vs
   url: http://current.geneontology.org/annotations/goa_human_rna.gaf.gz

PR #1184, opened 2019-09-26
adding _noiea to the dataset value in goa rna datasets

dougli1sqrd:


kltm:
As discussed in #1168 , we will be backing out of this.

kltm:
Will reuse.

dougli1sqrd:
Okay new commit. @kltm is this putting back the original IDs. Can you double check the `url` field to make sure that is correct?

kltm:
This looks fine to my eye.
As far as the `url` field goes...it's technically not up to date. See: https://github.com/geneontology/go-site/blob/b039a4b613d135f3f8a578f6adc5363b8cc9b69d/metadata/datasets/dictybase.yaml#L50
That said, I can't recall off the top of my head why we haven't updated these. Perhaps an oversight? Any thoughts @pgaudet ?
If this is what was historically in there, let's just go with that.

kltm:
@pgaudet Okay, this is weird; I'm going to reopen #804

PR #1189, opened 2019-10-03
Update group-contacts.csv

pgaudet:


metadata/group-contacts.csv
```diff
@@ -98,8 +98,8 @@ Rfam,Ioanna Kalvari,@kalvari,
 Rfam,Joanna Arganiska,@JoannaArgasinska,
 RGD,Stan Laulederkind,@slaulederkind,
 RGD,Tom Hayman,@gthayman,
-RGD,Jennifer Smith,jrsjrs,pipeline and data
-RGD,Marek Tuta,tutajm,import and export pipelines
+RGD,Jennifer Smith,@jrsjrs,pipeline and data
+RGD,Marek Tuta,tutajm,@import and export pipelines
```
kltm: @pgaudet I think you missed here...I don't think that's their username.

pgaudet: thanks !

PR #1190, opened 2019-10-04
Update to stats fieldnames & a few values

lpalbou:


lpalbou:
I have tested it locally and it works.

The commands are the same but for the first iteration the stats files for the previous releases are hosted in a different place, hence the PR: https://github.com/geneontology/pipeline/pull/137

kltm:
No affecting changes. Merge when you want.

PR #1197, opened 2019-10-10
changing the wording of gorule 33, and deprecating

dougli1sqrd:
For https://github.com/geneontology/go-site/issues/1082

### DIFF
--- a/metadata/rules/gorule-0000033.md
+++ b/metadata/rules/gorule-0000033.md
@@ -1,15 +1,15 @@
 ---
 layout: rule
 id: GORULE:0000033
-title: Public Reference IDs (PMID, PMC, doi, or GO_REF) should be preferred over group specific Reference IDs
+title: DEPRECATED. Public Reference IDs (PMID, PMC, doi, or GO_REF) should be preferred over group specific Reference IDs
 type: repair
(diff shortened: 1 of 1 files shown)


dougli1sqrd:
@pgaudet Here's the rewording that we discussed at the meeting. Let me know what you think?

kltm:
Pening discussion on https://github.com/geneontology/go-site/issues/1082

dougli1sqrd:
@pgaudet says we will deprecate this rule: #1082 (comment)
> After discussion with @ukemi and @vanaukenk
> We will deprecate this rule. Groups want to link to their internal pages for all references.

pgaudet:
@dougli1sqrd Can we look at this together ?

dougli1sqrd:
@pgaudet we can look at it together if you would like. But is my understanding wrong then when you said in the comment linked to above, that you and @vanaukenk and @ukemi have decided this rule should be deprecated? My thought was that we would then set the rule to be deprecated in the yaml, and turn off the processing of this rule.

pgaudet:
OK yes sure, if all you did is deprecate and inactivate the rule, then please merge. 
Thanks ! Pascale

PR #1200, opened 2019-10-15
change instance to input, like in the repair portion of the schema

dougli1sqrd:
The failing and passing example keys use `instance` currently, but the repair examples use `input` and `output`. Each example should use the same key name, and input seems apt.


kltm:
Makes sense

Issue #1202, opened 2019-10-24
Gaferencer should be run from the Mega Make for ontobio gaf parsing to make use of inferences


dougli1sqrd:
Gaferencer is now ready to be put in place in the go-site Makefile.

Running gaferencer on downloaded sources will produced a json file with data about reasoning from the GAF, including taxon check info. One thing that needs to happen so that gaferencer can be placed in the Makefile is have a runner script of sorts orchestrate the running and output so as to make gaferencer **group** specific, rather than gaf file specific.

The Makefile is oriented with a group view, as that is how the datasets yaml are oriented. Thus, gaferencer needs this as well.

This need was made evident in earlier testing for making the taxon checks work. There is an old version of this with a script: go-site/scripts/gaferencer_by_group.py in the branch `group_gaferencer_script` that ended up having everything but the kitchen sink. It became a testing ground. 

We can pull out the existing script there and the correct command line and redo this based on master now. I think it would be too messy to try and update master. There is a defunct PR https://github.com/geneontology/go-site/pull/1147 for this branch. 

This ultimately a part of https://github.com/geneontology/pipeline/issues/116 I would think.

dougli1sqrd:
As https://github.com/geneontology/go-site/pull/1203 was merged into master, and this was run on snapshot pipeline, this ticket is complete. There is an ongoing discussion about *results* of gaferencer https://github.com/geneontology/go-site/issues/1245.

PR #1203, opened 2019-10-24
metadata aware gaferencer script, calling gaferencer in makefile, passing in ontobio

dougli1sqrd:
For https://github.com/geneontology/go-site/issues/1202


kltm:
Already in `snapshot`, want to have go-site back on `master`.

PR #1213, opened 2019-11-05
Update db-xrefs.yaml to support sensible ECO

vanaukenk:
Adding information to ECO entry to allow linking out to the ECO site.

### DIFF
--- a/metadata/db-xrefs.yaml
+++ b/metadata/db-xrefs.yaml
@@ -672,7 +672,7 @@
     - type_name: entity
       type_id: BET:0000000
       id_syntax: \d{7}
-      url_syntax: http://www.evidenceontology.org/browse/#[example_id]
+      url_syntax: http://www.evidenceontology.org/browse/#ECO_[example_id]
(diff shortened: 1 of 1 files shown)


metadata/db-xrefs.yaml
```diff
@@ -664,15 +664,17 @@
...
   rdf_uri_prefix: http://purl.obolibrary.org/obo/ECO_
   generic_urls:
-    - http://www.geneontology.org/
+    - http://www.evidenceontology.org/
   entity_types:
     - type_name: entity
       type_id: BET:0000000
       id_syntax: \d{7}
-      example_id: ECO:0000002
+      url_syntax: http://www.evidenceontology.org/browse/#[example_id]
```
kltm: Should this not be "http://www.evidenceontology.org/browse/#ECO_"?

kltm: ^ @vanaukenk

vanaukenk:
@kltm - I think we're good now for ECO.
Merge?

Issue #1078, opened 2019-05-08
Update metadata for TAMU projects


jimhu-tamu:
A variety of files need to be updated and maintained for projects associated with the TAMU groups (@jimhu-tamu and @dsiegele). Following an email discussion with @vanaukenk and @kltm about whether we could do this directly via branching and pull requests is was suggested that I look at related documentation on 

http://wiki.geneontology.org/index.php/Installing_and_Using_git
http://wiki.geneontology.org/index.php/Ontology_Editors_Daily_Workflow

Based those, I am creating this Issue created to have a tracker id number for following the branch naming convention.

jimhu-tamu:
As I suspected, I can branch on my local machine, but cannot commit a new branch to go-site
```
git push --set-upstream origin issue-1078
ERROR: Permission to geneontology/go-site.git denied to jimhu-tamu.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

kltm:
For the time being, you can also fork and then create a PR back to us from your fork.

jimhu-tamu:
Created a pull request to test working from a fork. Doing it via the forked copy long-term is OK with me, assuming it works.

kltm:
More documentation now here:

https://github.com/geneontology/go-site/blob/master/documentation/how-do-i-edit-metadata.md

via https://github.com/geneontology/go-site/commit/8fb6640a0f4c2d74db42a5df0ad3dbc79c591bb2

kltm:
Is there any other metadata that you need changed?

jimhu-tamu:
Yes, I did a minimal change to test the workflow. 

The biggest thing is probably to manage where we will provide our post-svn gpad files (as we retire making gaf) and the stuff we used to email directly to Tony.

kltm:
@jimhu-tamu 
For the data stuff, the metadata in question for you would be located at: 
https://github.com/geneontology/go-site/blob/master/metadata/datasets/gonuts.yaml
https://github.com/geneontology/go-site/blob/master/metadata/datasets/ecocyc.yaml

We currently have GAF pickup (GPAD not quite there yet) and GPI pickup for loading into NEO (optional). If you need help setting up a place to drop files, please let us know.

jimhu-tamu:
I was just going to provide a url to a directory on our server. 

But does that mean we should still make gaf for now?

kltm:
[...]

PR #1215, opened 2019-11-07
Delete gonuts.yaml

kltm:
#1078


kltm:
@dougli1sqrd Any hard-codings for gonuts that would cause the pipeline to break?

kltm:
Does not seem to be loaded into AmiGO, so no change needed there.

dougli1sqrd:
No it should be fine. The only real place to look is if we are using it in the master Pipeline run as a test (I'm pretty sure we're not)

dougli1sqrd:
Indeed, I have double checked, and we are not.

kltm:
Cheers!

Issue #1219, opened 2019-11-12
gorule-report.html : some rules are not colored the right way according to whether they are filters or warnings


pgaudet:
For example gorule-0000046, gorule-0000008 are not colored.

dougli1sqrd:
This is now fixed here: https://github.com/geneontology/go-site/pull/1244

PR #1244, opened 2019-11-26
updating the coloring logic

dougli1sqrd:
This corrects the coloring logic in the gorules report as reported https://github.com/geneontology/go-site/issues/1219. 

What was happening is we would go through each dataset, and essentially reset the color status for each cell, clobbering any correct coloring. So if cycling through the data we happened to last leave on a dataset with no messages for a given rule, then we clobbered the color to be default. 

This has been fixed so that only higher level message colors can override lower level ones. (red is higher than yellow is higher than default).


kltm:
I'll say the logic looks good.

PR #1252, opened 2019-11-27
fixed the order in which we save off the exit status in rule example runner script

dougli1sqrd:


kltm:
This is more correct, if nothing else.

PR #1255, opened 2019-12-03
adding option in downloader that will unzip and zip all sources

dougli1sqrd:
so I know we don't strictly need this now, but I was most of the way done when SGN got back to us.

This adds an option to the downloader that will unzip all zipped, and zip all unzipped upstream sources as they are downloaded. This is what is then passed on to gaferencer and ontobio.

Accompanying Makefile change.


kltm:
Great--let's let this get through on `snapshot` after `release` finalizes.

kltm:
@dougli1sqrd Can we move on this now?

kltm:
Talking to @dougli1sqrd , this can be freely merged once current issues in the pipeline are worked out.

PR #1258, opened 2019-12-21
Delete pamgo

pgaudet:
@kltm I think you need to remove those sources from the pipeline code ?


pgaudet:
@kltm I let you do the merge ?

kltm:
@pgaudet In the future, it might be good to have a ticket/issue, and not just a PR. For now, I'll tag the other work onto here.

PR #1261, opened 2020-01-07
Update users.yaml

suzialeksander:
adding ShirinSaverimuttu per https://github.com/geneontology/helpdesk/issues/237


kltm:
Looks good; I'll update the production system a little later today.

PR #1264, opened 2020-01-07
Fixed go-rule to use a 'real' GAF line

pgaudet:
Previous one was truncated.


pgaudet:
@dougli1sqrd I don't understand what is not working now.

kltm:
Talking to @pgaudet , replaced by #1266

PR #1269, opened 2020-01-09
Added example for gorule-0000006

pgaudet:


pgaudet:
@dougli1sqrd @kltm Do you know why this is failing ?

kltm:
It looks like fundamental GAF parsing issues.
I'm guessing that you're using a symbol when you should be using an ID?

kltm:
Or count the number of fields again?

kltm:
Okay, it looks like you're missing the `with/from`--everything gets bumped back by one and the wheels fly off the unprotected parser.
I would suggest that you go ahead and make a "GAF template" in a spreadsheet, fill it in there, then export as TSV to drop in to these examples.

pgaudet:
@kltm I don't understand the error message ""success": false,

            "reason": "Input was expected to fail gorule-0000006, but it did not: extensions should be relation(curie)"" 

???

Thanks, Pascale

kltm:
@pgaudet Your fail line did not fail as expected.

pgaudet:
Yayyyyyyy !!!!

PR #1270, opened 2020-01-09
Changed schema to express deprecated resources

pgaudet:


pgaudet:
@kltm does this work to capture deprecated resources, ie adding to the schema: 
is_obsolete: boolean

kltm:
Obsolete in what context? Again, we are not the only ones that use this file. If a resource no longer has a usable CURIE -> URL/website map and it cannot be fixed, it should just be removed.

pgaudet:
OK - I was doing what @cmungall had suggested in the comments, but I also think it's cleaner if we remove the sources completely.

kltm:
I agree here, but if the mapping is legit I see no reason not to leave them.

pgaudet:
> I agree here, but if the mapping is legit I see no reason not to leave them.

Of course - when I say obsolete it means the URL is dead and I can find no new link to replace it. Most of the cases we know the resource is not maintained and its data not available anywhere. 

Should I add this 'obsolete' tag in the schema ? Or just delete the resource ?

kltm:
You should absolutely delete it: there is no utility in a mapping that doesn’t work.

PR #1325, opened 2020-01-17
Added gaf and gpad files for Xenbase.

malcolmfisher103:
This adds the requisite metadata for the GOC pipeline to pick up the annotation files produced by Xenbase. 

I hope this is all correct. I'm not sure if I should have added the destination URL for the 'current.geneontology.org' site, I was using the zfin.yaml file as a template.

There are some discrepancies between the contents of the Xenbase gaf and gpad files, due to issues converting noctua annotations to gaf format, but that shouldn't affect the metadata here at all and hopefully both files will be in better sync in the near future.

### DIFF
--- a/metadata/datasets/xenbase.yaml
+++ b/metadata/datasets/xenbase.yaml
@@ -7,51 +7,48 @@ project_url: "http://www.xenbase.org/"
 funding_source: "National Institute of Child Health and Human Development, grant P41 HD064556"
 email_report: "malcolm.fisher@cchmc.org"
 datasets:
- -
-   id: xenbase.gpi
-   label: "Xenbase gpi file"
-   description: "gpi file for Xenbase"
-   url: http://current.geneontology.org/annotations/xenbase.gpi.gz
-   type: gpi
-   dataset: xenbase
-   submitter: xenbase
-   compression: gzip
-   source: ftp://ftp.xenbase.org/pub/GenePageReports/xenbase.gpi.gz
-   entity_type:
-   status: active
-   species_code: Xenopus
-   taxa:
-    - NCBITaxon:8364
-    - NCBITaxon:8355
--
-   id: xenbase.gpad
-   label: "Xenbase gpad file"
-   description: "gpad file for Xenopus from Xenbase"
-   url: http://current.geneontology.org/annotations/xenbase.gpad.gz
-   type: gpad
-   dataset: xenbase
-   submitter: xenbase
-   compression: gzip
-   source: ftp://ftp.xenbase.org/pub/DataExchange/GO/xenbase.gpad.gz
-   entity_type:
-   status: active
-   species_code: Xenopus
-   taxa:
-    - NCBITaxon:8364
-    - NCBITaxon:8355
--
-   id: xenbase.gaf
-   label: "Xenbase gaf file"
-   description: "gaf file for Xenopus from Xenbase"
-   url: http://current.geneontology.org/annotations/xenbase.gaf.gz
-   type: gaf
-   dataset: xenbase
-   submitter: xenbase
-   compression: gzip
-   source: ftp://ftp.xenbase.org/pub/DataExchange/GO/xenbase.gaf.gz
-   entity_type:
-   status: active
-   species_code: Xenopus
-   taxa:
-    - NCBITaxon:8364
-    - NCBITaxon:8355
+  - id: xenbase.gpi
+    label: "Xenbase gpi file"
+    description: "gpi file for Xenbase"
+    url: http://current.geneontology.org/annotations/xenbase.gpi.gz
+    type: gpi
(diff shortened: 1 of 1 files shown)


kltm:
@malcolmfisher103 Can you confirm the taxon ids?

@pgaudet When adding, we'd want to have an analogous change to the pipeline (amigo loader) to pick up the generated GAF.

kltm:
@dougli1sqrd Would the GPI file be added for the entry if we are the ones generating it (from GAF I guess)?

kltm:
@dougli1sqrd Could you review the file at your convenience?

malcolmfisher103:
HI @kltm,

To confirm these files have annotations for two taxa 

NCBITaxon:8364 is Xenopus tropicalis
NCBITaxon:8355 is Xenopus laevis

malcolmfisher103:
> @dougli1sqrd Would the GPI file be added for the entry if we are the ones generating it (from GAF I guess)?

We at Xenbase generate a full gpi file for all genes on Xenbase, the same one we feed into NEO for use in Noctua. Does GOC generate a more specific one for annotation files?

malcolmfisher103:
From the Kwalify error for the xenbase.yaml file I seem to have left some spaces out of the lines dividing the different dataset sections. Is there a way for me to correct them with a further commit in this same pull request or does this pull need to be resolved first followed by a further commit to correct the missing spaces?

kltm:
@malcolmfisher103 It looks like your file is a bit off of the YAML spec. I have a corrected version, but cannot make direct changes as you are on a private fork. (In general, it's fine for consortium members to do PRs and branches within the org.)
I've added it here, if you could retry with it: https://gist.github.com/kltm/9e4b9abbd546209c6b765f9fb57f3cf7

kltm:
@malcolmfisher103 For your GPI question, if you are providing one that is great and all we need.

dougli1sqrd:
@kltm @malcolmfisher103 We do create our own GPIs from the upstream GAF, and that created GAF is what I assume is hosted on our site. There are open issues to tackle downloading and consuming upstream GPIs along with their GAFs.

The metadata file looks good!

kltm:
Almost there.
@pgaudet Is the intention for these to be added to the AmiGO (public) load immediately, or do you want to see them for a bit first?

pgaudet:
Should this go in the next release, so we can QC ? 

Just wondering though - how did the XenBase data used to be submitted?

kltm:
see https://github.com/geneontology/pipeline/issues/157

malcolmfisher103:
@pgaudet 
We have not previously been providing a data file, just making annotations directly in Noctua.

malcolmfisher103:
Thanks for the clarification @dougli1sqrd, that may mean there will be some genes covered in the GPAD not represented in the GPI as we are having some trouble integrating Noctua based annotations into the GAF.

malcolmfisher103:
I thought I should bump this as it is relevant to issue #1434 and it seems as if the pull never happened in the end. I'm not sure if the resolution of that issue will change which files I should be linking to in the 'xenbase.yaml' file.

kltm:
@malcolmfisher103 Could you describe a little about the contents of this GAF? As what you have PRed would make you the "canonical source" for these two species, you need to have already included any GOA sources (e.g. goa_uniprot_all), but no AmiGO/GO sources, as that would be eating our own tail.

malcolmfisher103:
@kltm Our current file does include AmiGO/GO sources. I know we were discussing this on the other thread and from my discussion with @ukemi last week I believe the idea was for us to provide a gaf file just with those goa_uniprot annotations and then the AmiGO and Noctua derived annotations would be added by your pipeline and we would pick up the complete file from some GOC endpoint. If so what files should these 'datasets' source files direct to? Our partial file that the GOC will ingest?

Currently the GOA files we are getting from EBI include annotations from a variety of sources including GOC and GO_Central. Some of these do seem to be duplicates, with small variation, of entries in AmiGO. Is there a specific set of sources we should exclude from the GOA files, or a set of sources to include, whichever is more appropriate? Please find below a table of the sources in the files we use from EBI, there is a count of the associated annotations as well.

|count|Source|
|----|----|
|62|AgBase|
|1|ARUK-UCL|
|26|BHF-UCL|
|3|DFLAT|
|5985|GOC|
|52|GO_Central|
|27|IntAct|
|71055|InterPro|
|1|MGI|
|1|ParkinsonsUK-UCL|
|3|Reactome|
|71731|UniProt|

kltm:
@malcolmfisher103 Yes, that is correct: we cannot be re-ingesting product from our previous runs like that--we'd need to have the "clean" file made available at a public location for pickup. If we go forward, PAINT and Noctua annotations would be added to the "clean" file for pickup by Xenbase.

kltm:
I believe that this is no longer current.

Issue #1380, opened 2020-02-12
Verbose mode and logging


dougli1sqrd:
Ontobio validate.py should have a "verbose" mode set with a flag.

This verbose mode will set logging to a more granular level.

The initial INFO logging we will do will pertain to tracking the bug here: https://github.com/geneontology/go-site/issues/1253

dougli1sqrd:
https://github.com/biolink/ontobio/pull/421

kltm:
This supports the pipeline overall.

PR #1381, opened 2020-02-12
adding -v on ontobio in the makefile to turn on verbose logging for now

dougli1sqrd:
https://github.com/geneontology/go-site/issues/1380


kltm:
Temporarily add flag to get at bug ASAP.

PR #1383, opened 2020-02-14
Add IAO to db-xrefs.

balhoff:
I am adding two IAO relations to the GO ontology ('editor note' and 'term tracker item') at @pgaudet's request. It seems like IAO needs to be added here for one of the ontology checks to pass.

### DIFF
--- a/metadata/db-xrefs.yaml
+++ b/metadata/db-xrefs.yaml
@@ -1161,6 +1161,12 @@
   entity_types:
     - type_name: entity
       type_id: BET:0000000
+- database: IAO
+  name: Information Artifact Ontology
+  description: Ontology of information entities, including ontology metadata relations.
+  rdf_uri_prefix: http://purl.obolibrary.org/obo/IAO_
+  generic_urls:
+    - http://purl.obolibrary.org/obo/iao
 - database: IMG
   name: Integrated Microbial Genomes; JGI web site for genome annotation
   generic_urls:


kltm:
@balhoff Huh. That a little odd--this is a file for mapping CURIEs to web pages. Can you point out that ticket or the mechanism?

balhoff:
It's used here in the GO makefile: https://github.com/geneontology/go-ontology/blob/79299ea5aafdd1fad735c9bf3bcbf86a77c2a044/src/ontology/Makefile#L151-L155

using this tool: https://github.com/geneontology/go-ontology/blob/master/src/util/check-obo-for-standard-release.pl

It's actually pulling from `https://s3.amazonaws.com/go-public/metadata/db-xrefs.legacy`. Would that be updated by updating this file?

kltm:
@balhoff It looks like you stumbled into something a little unpleasant. I've opened some tickets:

https://github.com/geneontology/pipeline/issues/162
https://github.com/geneontology/go-ontology/issues/18870

That said, fixing those will take a bit more work than we can tackle with this immediate issue.

While many of the ontologies in there apparently don't have it, it's always nice to have stanzas like the GO does:
``` yaml
  entity_types:
    - type_name: molecular_function
      type_id: GO:0003674
      id_syntax: \d{7}
      url_syntax: http://amigo.geneontology.org/amigo/term/GO:[example_id]
      example_id: GO:0004352
      example_url: http://amigo.geneontology.org/amigo/term/GO:0004352
```

That said, it's not in any way a blocker and you can merge at your leisure with or without.

Issue #1386, opened 2020-02-27
Error report - fix to remove OWL rules


pgaudet:
Hello, 

The top of the rules report includes a section about OWLTools checks that looks like this for http://snapshot.geneontology.org/reports/aspgd-report.html

> GAF Validation Summary Errors are reported first.

> For rule GO_AR:0000008 (http://www.geneontology.org/GO.annotation_qc.shtml#GO_AR:0000008) No annotations should be made to uninformative high level terms, there are 12415 violations with type Warning.

> For rule GO_AR:0000013 (http://www.geneontology.org/GO.annotation_qc.shtml#GO_AR:0000013) Taxon-appropriate annotation check, there are 243419 violations with type Warning.

I think those have all been superseded by other checks and this section can be removed. 

Also the  section it refers to (containing the actual errors for GO_AR:0000008 and GO_AR:0000013) should be removed, as they are redundant (and incorrect for GO_AR:0000013). 

Thanks, Pascale

dougli1sqrd:
http://skyhook.berkeleybop.org/issue-1386-remove-owltools-report/reports/goa_chicken-report.html

Owltools report is gone from the example pipeline run!

pgaudet:
Great !!!! Thanks @dougli1sqrd

dougli1sqrd:
This is in snapshot now! http://snapshot.geneontology.org/reports/fb-report.html

Closing.

pgaudet:
Great !!

PR #1391, opened 2020-02-27
removing owltools reports from merged group reports

dougli1sqrd:
For https://github.com/geneontology/go-site/issues/1386

Removing owltools reports in merged html file. The reports themselves will still be produced and still live, this just modifies how the reports are displayed in html.

As an aside, we should be aware of the version of the python `markdown` library. Current is 3.2.1, wherein the `markdown.extension.headerid` extension is deprecated. Our current version works, but we should consider changing how we produce the markdown -> html and update this extension to one that isn't deprecated, and then upgrade the markdown package.


kltm:
Removal of report and summary?

dougli1sqrd:
Yeah. I will still make a pipeline branch of this and run it, just so we see it w/out disruption first.

kltm:
Okay, sounds good.

dougli1sqrd:
Running in pipeline from a branch here: https://github.com/geneontology/pipeline/tree/issue-1386-remove-owltools-report

PR #1397, opened 2020-03-05
adding documentation in SOP about rule tags and context-import in rules schema

dougli1sqrd:


kltm:
Seems straightforward. Is this intended to be merged now, or as a separate branch for more work/testing?

dougli1sqrd:
This could be merged now. It'll have no mechanical affect. Those changes are in the ontobio PR: https://github.com/biolink/ontobio/pull/430

Even then, with the ontobio PR, there will be no difference. The main changes will occur once the hackathon branches start moving.

PR #1398, opened 2020-03-06
Hackathon 2020 rules

dougli1sqrd:
This is the work from the January Hackathon with @dustine32  and built upon since then. This mostly defines the two new rules we will be using in import.


dougli1sqrd:
This is now ready to merge.

kltm:
@dougli1sqrd It seems that the constraints TSV is being added here, instead of the preferred JSON?

dustine32:
@kltm The [JSON](https://github.com/geneontology/go-site/pull/1398/files#diff-3eb73beb831db23a37f3d1391d752bcc) is also added but hidden by the "large data don't display" shield. I don't think the TSV is actually used.

kltm:
@dustine32 Mind if I pull it from the PR then? We don't want to have two (possibly confusing) sources for the same info floating around.

dustine32:
@kltm Sure, I think pulling it should be fine.

dougli1sqrd:
@kltm @dustine32 just to confirm, we are keeping the YAML extensions-constraints, and removing the TSV?

I approve of that.

kltm:
So say we all.

I'm not sure if this is ready to go into master or not, but what I've seen makes sense and I'm fine with a merge if that's the desire (and I'll leave it to youseall).

dustine32:
Yep @dougli1sqrd correct! Keep the YAML (der! i keep thinking it's a JSON) and trash the TSV.

dougli1sqrd:
http://skyhook.berkeleybop.org/go-site-pr-1398-hackathon/reports/gorule-report.html

Pipeline finished which is good, but I feel like the new rules, rule 57 and 58, should not be run yet. Ontobio should not be running any import rules unless the import context is passed in, but here we are. So it looks like ontobio is actually misbehaving somehow.

I'll take a look.

kltm:
Half glass full!

dougli1sqrd:
Okay the new reports look better! http://skyhook.berkeleybop.org/go-site-pr-1398-hackathon/reports/gorule-report.html

57 and 58 are not running in a standard run of the pipeline.

kltm:
Is this to be merged then, or still working?

pgaudet:
Looks like it's still in progress ? Only some species are shown.

dougli1sqrd:
Yeah this is safely mergable at this point.

Issue #1413, opened 2020-03-12
Makefile and Dataset Metadata need to have the capability to turn on "import" mode for a pipeline run


dougli1sqrd:
For an "import" pipeline run, we need to let the Makefile and ontobio validation know to run the "import" rules in addition to the regular rule set.

The metadata schema should be changed to allow for a field, per dataset within each group to say if it's an `import` dataset or not.

Then, that data should be passed to ontobio via the makefile.

dougli1sqrd:
So, as ontobio already has the datasets metadata info in a validate run, we can determine whether a dataset is "import" or not, without a command line argument. Since this metadata would only happen in `validate`, we can safely universally look for the `import: true` in the datasets yaml on any run. This means we don't need a command line change in the makefile.

We will just add the needed schema changes in go-site, and then release a new ontobio with that capability.

dougli1sqrd:
http://skyhook.berkeleybop.org/go-site-1413-import_pipeline_test/reports/gorule-report.html

You can see that MGI has run rules 57 and 58, but no other dataset ran those rules.

With this, I call this a success, and we can update go-site to the latest ontobio.

PR #1421, opened 2020-03-16
updating schema to allow 'import' field to turn on import mode during…

dougli1sqrd:
… validation


dougli1sqrd:
For https://github.com/geneontology/go-site/issues/1413

kltm:
How to get there from here.

Issue #1420, opened 2020-03-16
gorule-0000030 - Needs to be a warning


pgaudet:
There are too many annotations that fails. Can you change this ? We need this for the next release to go out. 

Thanks, Pascale

kltm:
@pgaudet I'm trying to understand the status change here. Is there a bug in 0000030? We've had that as it is now for well over a year and only threee species seem to be having trouble. Is there a reason for the sudden change?

pgaudet:
Yes, I have been cleaning up redundant gorefs and some groups annotations
to now obsoleted gorefs.

We should keep it a warning until the gorefs are completely cleaned up.

BTW it would be useful if the Readme. MD page was updated.

Thanks, Pascale

kltm:
@pgaudet 
"We should keep it a warning until the gorefs are completely cleaned up."
Could you make a ticket for the re-reversion then? We can reference that when we close out this one.

kltm:
@pgaudet That said, is switching to "warning" (and letting thousands more annotations in) actually blocking a `release` run for you?

pgaudet:
Yes @kltm this is OK. 
In the current release, this rule has no errors. We are not introducing any errors in the data, we just want people to switch to a different reference number. 

Thanks for confirming, 

Pascale

kltm:
@pgaudet Apologies, I'm still unclear with the response in https://github.com/geneontology/go-site/issues/1420#issuecomment-599929378. Is turning this rule to a warning actually blocking requesting a `release` run for you?

- [ ] yes
- [ ] no

As well, could you make a ticket for the re-reversion (switching this rule back to an "error" from a "warning")?

pgaudet:
This is why I put 'blocker' as the label. 

Thanks, Pascale

dougli1sqrd:
Okay, let's check back in the next snapshot build. I've updated ontobio with the new gorule-0000030 implementation along with changing the rule document in go-site. Ontobio is released and the new version with the change is set in go-site.

kltm:
@dougli1sqrd Has this been cleared?

dougli1sqrd:
It looks like it. I believe we can close this issue.

kltm:
I can confirm that gorule-0000030 appears as a warning in the last `snapshot` report.

PR #1424, opened 2020-03-17
updating rule 30 to be a warn/report instead of error

dougli1sqrd:
For https://github.com/geneontology/go-site/issues/1420


kltm:
Looks good.

PR #1425, opened 2020-03-18
adding shex reports page gen and template

dougli1sqrd:
This is a script that folds a json that looks like:

```
{
  "taxon": "taxon unknown",
  "number_of_models": 12,
  "number_of_models_in_error": 0,
  "number_of_correct_models": 0,
  "messages": {
    "gorule-0000019": [
      {
        "level": "ERROR",
        "model-id": "gomodel:5b528b1100000186",
        "type": "Violates GO Rule",
        "obj": "",
        "taxon": "taxon unknown",
        "message": "GORULE:0000019: Model is logically inconsistent",
        "rule": 19
      },
...
```

into a report html that mirrors the gorule report page.

Report screenshot attached as an example
Screenshot from 2020-03-18 14-24-39

**Note!** These hyperlinks won't go anywhere, and so is currently misleading. This is just spending the minimum time making this work with the reports.
.

### DIFF
--- a/scripts/shex-reports-page-gen.py
+++ b/scripts/shex-reports-page-gen.py
@@ -1,0 +1,175 @@
+# python3 ./scripts/shex-reports-page-gen.py --report ./pipeline-report.json --template ./scripts/shex-reports-page-template.html --date 2019-03-26 > shex-report.html
+
+import click
+import pystache
+import json
+import datetime
+import yamldown
+import os
+import glob
+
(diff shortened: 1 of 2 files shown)


dougli1sqrd:
For https://github.com/geneontology/pipeline/issues/167 and see also https://github.com/geneontology/minerva/issues/254

kltm:
Talked to @dougli1sqrd , looks good.

PR #1447, opened 2020-04-03
Update users.yaml

hdrabkin:
adding additional users for MGI to aid in annotation imports

### DIFF
--- a/metadata/users.yaml
+++ b/metadata/users.yaml
@@ -4122,3 +4122,18 @@
   nickname: 'Gary Schindelman'
   organization: WB
   uri: 'http://orcid.org/0000-0001-7883-4388'
+-
+  nickname: 'Robert Sinclair'
+  organization: MGI
+  uri: 'GOC:bobs'
+  xref: 'GOC:bobs'
+ -
+  nickname: 'Louise McKenzie'
+  organization: MGI
+  uri: 'GOC:lmm'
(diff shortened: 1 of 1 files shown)


pgaudet:
@vanaukenk Should we make it a policy that all active users in users.yaml also ned to be in groups-contacts.cvs ?

ukemi:
Doe that make sense? I thought that the contacts were supposed to be the point person or at most a couple of people at each group, not everyone active in a project.

vanaukenk:
@pgaudet @ukemi 

I think this depends on what the intended purpose of the groups-contacts.csv file is (or what we want it to be).

If we want it to serve as the source of information for groups, names, and github handles of anyone active in the project, that's a good place to collect that info.  It's certainly easier to find it there than scanning or searching through the users.yaml.

Here's a suggestion: we include everyone, but also designate one or two point people for each project in a separate column.  And perhaps consider retiring the expertise column.  It seemed like a good idea at the time, but is incomplete or mostly stating the obvious, i.e. curators at MODs are the experts for that organism.

kltm:
@pgaudet https://github.com/geneontology/go-site/pull/1447#issuecomment-609637840 would obviate the need for groups-contacts.cvs, as everything could be in users.yaml. My recollection agrees with https://github.com/geneontology/go-site/pull/1447#issuecomment-609750113 -- these had separate uses.

Issue #1443, opened 2020-04-01
Ensure that GPAD (2.0) can be used as a primary input into the pipeline


kltm:
As we move forward, with the active Noctua imports being the deadline, the pipeline needs to be able to process GPAD 2.0 as a primary input component (GAF being the only other primary data component at this time).

Tied to this, to actually make use of GPAD elsewhere in the pipeline (e.g. produce GAF for AmiGO to consume), we'll also need to be making use of GPI 2.0 files.

Tagging @dougli1sqrd

dougli1sqrd:
Datasets Yaml proposal:
GPAD datasets should have a new field `gpi` that is a list of dataset IDs within the group that refers to GPI files that are required for download and processing along with the gpad:
```
id: foo.gpad
dataset: foo
gpi:
  - foo.gpi
```

One question though that I have is which datasets are we considering to by the "correct" canonical datasets for a group? Now that we are expecting multiple types of files (gaf or gpad) we need a way of deciding which one is more canonical. 

For example, in the `goa.yaml` group, we have 3 entries that have `dataset: goa_chicken`: `goa_chicken.gaf`, `goa_chicken.gpad`, `goa_chicken.gpi`. This essentially represents 2 actual sources, as gpad+gpi is one, and then gaf. So even if we linked gpad to it's gpi, we still have to know which to use as the canonical.

We could just go totally one to one: where the goa_chicken.gaf gets processed to our canonical goa_chicken.gaf, the goa_chicken.gpad/gpi gets processed into our goa_chicken.gpad/gpi. To contrast, our current process takes GAF, and then produces all types from the one file. So we have a one to many.

We could also indicate in the yaml file which datasets are the blessed ones to use. Like:
```
# In goa.yaml
id: goa
sources:
  - goa_chicken.gpad
  - goa_chicken_isoform.gpad
  - goa_human.gpad
  - goa_cow.gaf
  ...
datasets:
  - id: goa_chicken.gpad
    gpi: [goa_chicken.gpi]
```

This allows the providers to say that *this* gaf source should be used, while *that* gpad source should be used.

Or perhaps there should be a flag in the datasets stanzas?

Ultimately, we need a way to map to our outputs. And having 3 sources that have `dataset: foo` in a file is unclear what to use. What do we think?

kltm:
[...]

PR #1455, opened 2020-04-08
WIP Support for GPAD as a primary product in the pipeline

dougli1sqrd:
For https://github.com/geneontology/go-site/issues/1443


kltm:
@dougli1sqrd , I think we probably want to have a quick chat with @balhoff and maybe @goodb about overall pipeline goals and how GPI files are produced from Noctua (and how that fits into the overall data flow).

Issue #1431, opened 2020-03-23
Add available coronavirus data to the pipeline


kltm:
Add available SARS-CoV-2 data to the pipeline

- [x] @alexsign to produce GPI file
- [x] add to NEO (change config line in Makefile)
- [x] @alexsign to produce GAF/GPAD (this will be mostly interpro2go etc to start with)
- [ ] Add GAF/GPAD to yaml, so can be loaded into amigo, added to release files
- [ ] Patrick/ViralZone will do a GO-CAM for SARS-CoV-2
- [ ] This should naturally flow into GO-CAM site. @lpalbou look into a way to highlight
- [ ] @alexsign to load GPADs emanating from GO-CAMs back into GOA
 - [ ] **UPDATE 2020-05-13** do the same for SARS-CoV

Tagging @pgaudet @cmungall 

Questions:

 - @alexsign is it easy for you to give us the GPI in advance of the proteins going in to uniprot main release? If not, it is trivial for us to parse the xml from ftp://ftp.uniprot.org/pub/databases/uniprot/pre_release/
 - @alexsign will the GPI include all of the isoforms? It looks from coronavirus.xml on the EBI FTP site at the moment there is only accessions for the GPCR proteins

pgaudet:
@kltm you want a GPI file with only the one viral species ?

cmungall:
We previously discussed just the SARS-CoV-2 genome, but we could extend to the other coronavirus genomes. But let's do the SARS-CoV-2 genome first

alexsign:
ftp://ftp.ebi.ac.uk/pub/contrib/goa/uniprot_reviewed_sars-cov-2.gpi
ftp://ftp.ebi.ac.uk/pub/contrib/goa/uniprot_sars-cov-2.gpi

cmungall:
Great! @thomaspd indicated we may need the isoforms also - is this something that will require further upstream protein sequence curation in uniprot?

alexsign:
@cmungall I don't think we have this data in uniprot private release yet.

alexsign:
ftp://ftp.ebi.ac.uk/pub/contrib/goa/uniprot_sars-cov-2.gpa

kltm:
@cmungall Your neo changes are failing with: 
```
18:18:26  make: *** No rule to make target 'target/neo-goa_sars-cov-2.obo', needed by 'all_obo'.  Stop.
```

alexsign:
[...]

PR #1459, opened 2020-04-09
WIP addition of goa_sars-cov-2 metadata for #1431

kltm:
Still needing to wait on https://github.com/geneontology/go-site/issues/1431#issuecomment-611799993 for the moment.

Would like a check on these fields:

```yaml
   entity_type: rna
   species_code: Scov2
   taxa:
    - NCBITaxon:694009
```

@cmungall

### DIFF
--- a/metadata/datasets/goa.yaml
+++ b/metadata/datasets/goa.yaml
@@ -1056,6 +1056,22 @@ datasets:
     - NCBITaxon:9823
 
  -
+   id: goa_sars-cov-2.gaf
+   label: "goa_sars-cov-2 gaf file"
+   description: "gaf file for goa_sars-cov-2 from EBI Gene Ontology Annotation Database"
+   url: http://geneontology.org/gene-associations/goa_sars-cov-2.gaf.gz
+   type: gaf
+   dataset: goa_sars-cov-2
+   submitter: goa
+   compression: gzip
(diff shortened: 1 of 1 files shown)


cmungall:
Entity type should be protein (Yes it is an rna virus, but the entries are proteins. I'm not sure it matters as this field doesn't seem to be used anywhere?)

aside: The entity_type field seems to be undocumented. we really need to use something standard like json-schema for the metadata schemas.

I chose Scov2 arbitrarily as the 'species code'. I don't know if uniprot have assigned an official one yet but I think 'Sars2' may be better -- not sure this matters either.

kltm:
The last iteration appears to have been `SARS-CoV`; I haven't been able to find something more up to date.

kltm:
Actually this may be the right one:
https://www.uniprot.org/docs/speclist
```
CVHSA V  694009: N=Human SARS coronavirus
                 C=SARS-CoV
```

kltm:
NM: `SARS-CoV-2`...I'm still having trouble finding a taxon ID. I'll try again later.

kltm:
Sorry: https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?id=2697049

kltm:
Now waiting on clearing: https://github.com/geneontology/go-site/issues/1431#issuecomment-611799993

Issue #1206, opened 2019-10-31
Update URLs for Xenbase


vanaukenk:
The current crop of URLs in the dbxefs.yaml file for Xenbase do not appear to be working.

@malcolmfisher103 may be able to help us get the correct ones?

We'd like to have link outs from the Noctua Form 2.0 autocomplete to Xenbase for Xenbase gene IDs.

malcolmfisher103:
I'll look into this. 

Nothing looks obviously wrong at first glance, but maybe I am misinterpreting how the various fields are being used.

Is the autocomplete using the ID syntax regexes to get the IDs and substituting those into the given 'url_syntax:' format to generate appropriate URLS? Are the 'rdf_uri_prefix:' fields being used at all?

vanaukenk:
Thanks, @malcolmfisher103 
We've started another ticket in the amigo tracker to try to resolve this issue.
We'll let you know if we need any more help from Xenbase.

malcolmfisher103:
Great.

Just FYI, I have discovered some irregularities with how the given URLs resolve currently - 

The http://www.xenbase.org/common/xsearch.do?searchValue=[example_id] form is lacking some restrictions that can lead to issues. 

With the restrictions added the form should be -
http://www.xenbase.org/common/xsearch.do?exactSearch=true&searchIn=7&searchValue=[example_id].

kltm:
Now:
http://www.xenbase.org/common/xsearch.do?exactSearch=true&searchIn=7&searchValue=XB-GENE-6255962
http://www.xenbase.org/common/xsearch.do?exactSearch=true&searchIn=7&searchValue=[example_id]

kltm:
https://github.com/geneontology/amigo/issues/581#issuecomment-548996135

malcolmfisher103:
Hi Seth, Any idea what is causing the problem? The re-formatted links work just using basic substitution. Does anything get parsed out by the processes ingesting the yaml that might conflict with the messy search query nature of our links? There is an identifiers.org referrer we could try instead, it has the syntax 'https://identifiers.org/xenbase:XB-GENE-[example_id]'. I had been avoiding it as it only works for genes not any of our other bioentities so our link syntaxes in the 'dbxref.yaml' file would be less consistent.

ukemi:
Hmmmm. I’m in. Try a complete refresh.

kltm:
[...]

PR #1468, opened 2020-04-18
Updated Xenbase entity id and url syntaxes.

malcolmfisher103:
Changed all Xenbase entity type url_syntax entries to 'http://www.xenbase.org/common/xsearch.do?exactSearch=true&searchIn=7&searchValue=[example_id]' and id_syntax entries to 'XB-(GENE|LINE|TRANSGENE|MORPHOLINO)-[0-9]+'. In line with discussion in #1206.


kltm:
@malcolmfisher103 Cheers--this looks like what we discussed.

PR #1492, opened 2020-05-14
Add generic gramene ensembl dbxref

elserj:
This one we'd like to test sooner then the other ones I mentioned as it may supersede those needing to be added.


kltm:
@elserj The example URL doesn't seem to go anywhere?

elserj:
D'oh. Missed the question mark in the links. This better?

kltm:
This seems right to me.

PR #1504, opened 2020-06-08
add script to replace sh commands in pipeline gaference reprocessing

dougli1sqrd:
This will just replace lines 680 to 697 in the current master Jenkinsfile.


dougli1sqrd:
The build is failing because of a previous commit to db-xrefs.yaml that fails the schema. I'm going to merge this.

Issue #1484, opened 2020-05-06
Some taxon contraints still incorrect


pgaudet:
Hello, 

In this report: 
http://release.geneontology.org/2020-05-02/reports/goa_chicken-report.html#gorule-0000013

This line is a false positive: 
ERROR - Violates GO Rule: GORULE:0000013: Taxon-appropriate annotation check -- UniProtKB O93574 RELN GO:0021800 GO_REF:0000024 ISS UniProtKB:Q60841 P Reelin RELN protein taxon:9031 20091116 UniProt

GO:0021800 taxon constraints should allow this annotation in Chicken. 

Can you please check what's going on ? 

Thanks, Pascale

dougli1sqrd:
Jim says that it's because Uberon says that 'corpus callosum' in only in Eutheria (mammals)
(@balhoff)

https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2666078/

balhoff:
@dougli1sqrd @pgaudet I looked at a different line actually (GO:0021540 'corpus callosum morphogenesis').

For this one it is actually because neocortex is only in Mammalia.

pgaudet:
Thanks. Is there any way accessing taxon constraints for each term can be made more user friendly ? If they were displayed in AmiGO it'd be great. Right now it's hard to request people to review their annotations if finding the issue is so difficult. 

Thanks, Pascale

@cmungall

balhoff:
@pgaudet we have talked about a couple options:
- Our tool gaferencer has an option to output a matrix of GO term by taxon, indicating whether the term is always in, never in, or unstated. We can add that to the pipeline somewhere and make it available as a product. It's a large matrix.
- We can add explanation output to the part of gaferencer that @dougli1sqrd uses to generate the reports you're looking at in this ticket. So the rule violation could include all the axioms, or maybe just the in_taxon axiom. The problem with this is that computing each explanation takes anywhere from several seconds to a couple minutes. This would be okay if there weren't too many taxon violations.

kltm:
Okay, after a conversations about this today we have a basic plan in place. Currently, I believe this is what we have:

- [ ] preserve the gafrencer products, tarballed and gzipped, from the mega-step step in skyhook (/products/gafrencer/*) (check spelling)
- [ ] create a *temporary* new stage just after the mega-step, based on the mega-step to preserve the same environment (i.e. https://github.com/geneontology/pipeline/blob/master/Jenkinsfile#L499-L505)
[...]

PR #1509, opened 2020-06-10
Moving pigz and all.gaferences.json into make; for geneontology/go-si…

dustine32:
@kltm @dougli1sqrd For https://github.com/geneontology/go-site/issues/1484.

Note: I believe `BASE_NAMES = $(notdir $(wildcard /opt/go-site/annotations/*))` should now be picking up files with `.gz` extensions since extraction is now inside make.

Though I think the wildcard will also pick up the `*.gpi.gz` files. Should we have a recipe for these or exclude them in the wildcard?

### DIFF
--- a/scripts/Makefile-gaf-reprocess
+++ b/scripts/Makefile-gaf-reprocess
@@ -1,9 +1,22 @@
 BASE_NAMES = $(notdir $(wildcard /opt/go-site/annotations/*))
 
-all: $(foreach name,$(BASE_NAMES),/opt/go-site/annotations_new/$(name))
+all: $(foreach name,$(BASE_NAMES),/opt/go-site/annotations_new/$(name)) /opt/go-site/gaferencer-products/all.gaferences.json.gz
 
-/opt/go-site/annotations_new/%.gaf: /opt/go-site/annotations/%.gaf
+/opt/go-site/annotations_new/%.gz: /opt/go-site/annotations_new/%
+	pigz /opt/go-site/annotations_new/$*
+
+/opt/go-site/annotations_new/%.gaf: /opt/go-site/annotations/%.gaf /opt/go-site/gaferencer-products/all.gaferences.json
 	ontobio-parse-assocs.py -f $< -F gaf -o $@ -I /opt/go-site/gaferencer-products/all.gaferences.json --report-md /tmp/report.md --report-json /tmp/report.json convert -to gaf
 
-/opt/go-site/annotations_new/%.gpad: /opt/go-site/annotations/%.gpad
+/opt/go-site/annotations_new/%.gpad: /opt/go-site/annotations/%.gpad /opt/go-site/gaferencer-products/all.gaferences.json
 	ontobio-parse-assocs.py -f $< -F gpad -o $@ -I /opt/go-site/gaferencer-products/all.gaferences.json --report-md /tmp/report.md --report-json /tmp/report.json convert -to gpad
+
+/opt/go-site/annotations/%:
+	unpigz /opt/go-site/annotations/$*.gz
+
+/opt/go-site/gaferencer-products/all.gaferences.json.gz: /opt/go-site/gaferencer-products/all.gaferences.json
+	pigz /opt/go-site/gaferencer-products/all.gaferences.json
+
+/opt/go-site/gaferencer-products/all.gaferences.json:
+	tar -xvf /opt/go-site/gaferencer-products/gaferences.json.tgz -C /opt/go-site/gaferencer-products
+	python3 /opt/go-site/scripts/json-concat-lists.py /opt/go-site/gaferencer-products/*.gaferences.json /opt/go-site/gaferencer-products/all.gaferences.json


scripts/Makefile-gaf-reprocess
```diff
@@ -1,9 +1,22 @@
 BASE_NAMES = $(notdir $(wildcard /opt/go-site/annotations/*))
 
-all: $(foreach name,$(BASE_NAMES),/opt/go-site/annotations_new/$(name))
+all: $(foreach name,$(BASE_NAMES),/opt/go-site/annotations_new/$(name)) /opt/go-site/gaferencer-products/all.gaferences.json.gz
 
-/opt/go-site/annotations_new/%.gaf: /opt/go-site/annotations/%.gaf
+/opt/go-site/annotations_new/%.gz: /opt/go-site/annotations_new/%
+	pigz /opt/go-site/annotations_new/$*
```
dougli1sqrd, line 6: I'd recommend using `$<` maybe instead of trying to reconstruct the path, when you already have it in the prereq.

scripts/Makefile-gaf-reprocess
```diff
@@ -1,9 +1,22 @@
...
 
-/opt/go-site/annotations_new/%.gpad: /opt/go-site/annotations/%.gpad
+/opt/go-site/annotations_new/%.gpad: /opt/go-site/annotations/%.gpad /opt/go-site/gaferencer-products/all.gaferences.json
 	ontobio-parse-assocs.py -f $< -F gpad -o $@ -I /opt/go-site/gaferencer-products/all.gaferences.json --report-md /tmp/report.md --report-json /tmp/report.json convert -to gpad
+
+/opt/go-site/annotations/%:
+	unpigz /opt/go-site/annotations/$*.gz
+
+/opt/go-site/gaferencer-products/all.gaferences.json.gz: /opt/go-site/gaferencer-products/all.gaferences.json
+	pigz /opt/go-site/gaferencer-products/all.gaferences.json
```
dougli1sqrd, line 18: I think you may be able to roll all `pigz` and `unpigz` steps into only 2 recipes, on to unzip, and one to zip. 
If every time we zip/unzip the product goes into the same directory, we can make a generic recipe:

```
%.gz: %
    unpigz $<

%: %.gz
    pigz $<

I think that should work?
```

scripts/Makefile-gaf-reprocess
```diff
@@ -1,9 +1,22 @@
...
+
+/opt/go-site/gaferencer-products/all.gaferences.json:
+	tar -xvf /opt/go-site/gaferencer-products/gaferences.json.tgz -C /opt/go-site/gaferencer-products
```
dougli1sqrd, line 21: I feel confused about this. This is implying that presumably we downloaded a `all.gaferences.json.tgz` from skyhook, and that now we're appending to it? Why do we have an `all.gaferences.json.tgz` if doesn't already include all the gaferences from the products? Does the mega-make step upload the individual gaferences and bundle them together? Or does the bundling only happen here? In which case we shouldn't have acces to a `gaferences.tgz` yet, right?

scripts/Makefile-gaf-reprocess
dustine32: Genius!

scripts/Makefile-gaf-reprocess
dustine32: Referring to https://github.com/geneontology/go-site/issues/1484#issuecomment-639177452:
> preserve the gafrencer products, tarballed and gzipped, from the mega-step step in skyhook (/products/gaferencer/*)

This tarball is `gaferences.json.tgz`, generated and pushed to skyhook in the mega-make step. And then
> pull in gafrencer tarball and convert those files into a single JSON file

This converted single JSON is `gaferences.json.gz` (aka `all.gaferences.json.gz` in the pipeline). This `all.gaferences.json` is what's used by `ontobio-parse-assocs.py`, right?

dustine32: Oh yeah, and `gaferences.json.tgz` is rsync'd in from skyhook (along with `annotations/*`) right before the `make` cmd:
https://github.com/geneontology/pipeline/blob/f8dc05dc48e7fa9c1f7e55eb1f25f5e49a72a5da/Jenkinsfile#L675-L676
So it should be accessible by `make`.

dougli1sqrd: Okay just so I have this straight: so when we download, all.gaferences.json.tgz is the bag of all the different gaferencer products from the megam-make step. Then when unzipped, it pops out a bunch files that are the gaferences output from the metagstep, unzipped. Then we concatenate into all.gaferences.json, to be used in the ontobio command. Then we zip all.gaferences.json into all.gaferences.json.tgz, which only includes the one mega gaferences file? 

Is there a reason we shouldn't/couldn't do the concat *in* the mega-make? And then this step just has to download the one gaferences file essentially?

dustine32: @dougli1sqrd Oh sure, I think it can be concatenated in the mega-make step. I'll try it out!

And right about everything else! It does sound pretty convoluted, doesn't it.

PR #1512, opened 2020-06-11
fixing makefile to be testable locally and to stop circular rules

dougli1sqrd:


dougli1sqrd:
To run locally, if you download the files to go-site/annotations, and the gaferencer file to gaferencer-products, you can run with
```
make -f scripts/Makefile-gaf-reprocess ROOT_PATH=$(pwd) all
```

By default ROOT_PATH is /opt/go-site (where it will be on docker)

Issue #1514, opened 2020-06-11
Clarify the ability for basic pipeline tooling to parse and produce UTF-8 in files


kltm:
With the recent discussion on whether or not to include UTF-8 in the GPAD/GPI 2.0 spec, we feel that it is important to know if the basic parser and emitter functionality supports UTF-8 already, and if not how much additional work it might be to support it.

Closing this ticket would include clarifying basic clarifying information for GAF input and output. Most likely using go rule 0000001 GAF line i/o checks as a proxy.

PR #1515, opened 2020-06-11
Add basic GAF pass-through check for UTF-8

kltm:
#1514


kltm:
@dougli1sqrd @balhoff 
Just confirming that this passing demonstrates that the current base GAF parser and emitter likely allows UTF-8 characters from a few categories to pass through.

PR #1548, opened 2020-09-08
Update users.yaml

suzialeksander:
just making sure uri should be http not https (see lines 4166, 4178)


kltm:
@suzialeksander Not to derail this, but do you have a YAML editor? There seems to be a couple of syntax mistakes in there...

suzialeksander:
No, I'm copy/pasting as usual.  I have something that will edit in YAML (external editor, BBEdit, it's showing me pretty colours but I don't know if it checks formatting) , but I don't think my copy/paste skills have failed this badly before.

Issue #1208, opened 2019-11-01
genedb_pfalciparum GAF file


pgaudet:
Hi @Achchuthan 

We will migrate away from the GO SVN in the upcoming months. We still get the genedb_pfalciparum GAF file from the GO SVN (http://viewvc.geneontology.org/viewvc/GO-SVN/trunk/gene-associations/submission/gene_association.GeneDB_Pfalciparum.gz)

Would it be possible to move the file that you produce to the same location on the Sanger FTP as other GeneBD files ? see 
The https://github.com/geneontology/go-site/blob/master/metadata/datasets/genedb.yaml

Thanks, Pascale

pgaudet:
@Achchuthan Did you ever have a change to look into this ? The SVN will soon be retired.

pgaudet:
OK got it - this is provided by GOA in the UniProt-all file. I'll update the genedb yaml accordingly. 

Thanks,

Pascale


@kltm

kltm:
See https://github.com/geneontology/go-site/pull/1555

Achchuthan:
Dear @pgaudet,

Sorry for the late reply to your comment from 4 days ago due to sick child and leave last week. I have somehow missed this issue from Nov 2019. I'll ask @Uliboehme and the GeneDB informatics team to push the file to the FTP site. We can add Plasmodium falciparum back to GOA after they generate this.

Thanks & best wishes,
Achchuthan

pgaudet:
@Achchuthan Thanks for the reply. I hope everyone is doing well now. We were planning to use GOA for the upcoming release (October 2020) since GOA has more annotations from Plasmodium than the file on the svn does. 
Please let us know if we should not do that. 


NCBITaxon:36329 Plasmodium falciparum 3D7 => GOA: 33710 ; AmiGO => 14813

NCBITaxon:5833 Plasmodium falciparum =>  GOA: 99676 ; AmiGO =>  0

Thanks , Pascale

Achchuthan:
Hi @pgaudet, 
Thank you very much. I think using GOA should be fine for October release. We can internally compare the differences between GOA and our updated annotations and come up with a solution for the future releases after the October one. @Uliboehme is the Plasmodium curator and I have forwarded this issue to her.

Thanks,
Achchuthan

PR #1555, opened 2020-09-18
Update genedb.yaml

pgaudet:


kltm:
It's good to tie these to their ticket.
Adding https://github.com/geneontology/go-site/issues/1208

PR #1559, opened 2020-09-24
Update users.yaml

vanaukenk:
Trying to add Cristina


kltm:
@vanaukenk Just checking if the users were refreshed for noctua prod/dev? If we need that done, you can also ping me.

pgaudet:
Is the procedure written anywhere for this ? It would be nice to have all the steps explicit somewhere.

vanaukenk:
@kltm 
Not yet; I'll do it this morning.

Issue #1522, opened 2020-06-24
GOA to generate new GAF/GPAD/GPI files 'go_references_species'


pgaudet:
In addition to goa_all, GOA will generate new GAF/GPAD/GPI files 'go_references_species' that correspond to the 142 species from Panther (and the substrains corresponding to those species). 

List of species is in the attached file. [
[Pthr15_proteome_UPs.txt](https://github.com/geneontology/go-site/files/4825203/Pthr15_proteome_UPs.txt)
](url)

dustine32:
I've got the PANTHER 15.0 species containing just the 142 "leaf" species and omitting the ancestor species. It's now on our data server here http://data.pantherdb.org/PANTHER15.0/globals/species_list.tsv and will be produced with each new PANTHER version. I will also commit this TSV to the `go-site/metadata` folder as `go_reference_species_list.tsv`.

mugitty:
These are also available via API (http://pantherdb.org/services/oai/pantherdb/supportedgenomes)

PR #1560, opened 2020-09-24
Added list of 142 species in PANTHER 15.0 for #1522

dustine32:
Here's the list in TSV format with potentially too many columns. Let me know if you need this file changed (like TSV->YAML or more/fewer columns) before merging.


dustine32:
Referencing #1522

PR #1561, opened 2020-10-01
update ontobio to 2.1.0 (mostly bug fixes in gpad parser)

dougli1sqrd:


kltm:
I'm guessing it's unlikely to break standard things?

PR #1583, opened 2020-11-18
add PomBase GPAD & GPI entries

mah11:
Note: Source URLs do not include ".gz", because @kimrutherford says that https connections to www.pombase.org are automatically compressed.

### DIFF
--- a/metadata/datasets/pombase.yaml
+++ b/metadata/datasets/pombase.yaml
@@ -31,7 +31,6 @@ datasets:
    type: gaf
    dataset: pombase
    submitter: pombase
-   compression: gzip
    source: https://www.pombase.org/data/annotations/Gene_ontology/pombase.gpi
    entity_type:
    status: active
@@ -47,7 +46,6 @@ datasets:
    type: gaf
    dataset: pombase
    submitter: pombase
-   compression: gzip
    source: https://www.pombase.org/data/annotations/Gene_ontology/pombase.gpad
    entity_type:
    status: active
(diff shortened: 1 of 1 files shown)


kltm:
@dougli1sqrd Can we currently accept non-compressed sources?

kimrutherford:
> Note: Source URLs do not include ".gz", because @kimrutherford says that https connections to [www.pombase.org](http://www.pombase.org) are automatically compressed.

Just to be clear, the client will need to request it.  For example with curl the extra header needs to be added like:

```
curl -O -H 'Accept-Encoding: gzip' 'https://www.pombase.org/data/annotations/Gene_ontology/pombase.gpi'
```

> @dougli1sqrd Can we currently accept non-compressed sources?

We can provide .gz versions if needed.

pgaudet:
@dougli1sqrd I think @kimrutherford is waiting for you to say whether you need the zipped file ? 
Or is something else preventing this to be merged ?

dougli1sqrd:
We can accept `gzip` or uncompressed files.

However, if you're supplying uncompressed then the `compression` field should be empty, or that field should be removed. If the pipeline sees that a dataset has `compression: gzip` then it will attempt to unzip it thinking it is compressed.

mah11:
OK, let's remove the `compression: gzip` lines. How do we do that? Should I do another commit on the branch with the pombase changes (patch-14)?

kimrutherford:
> How do we do that? Should I do another commit on the branch with the pombase changes (patch-14)?

Yep, that should work.

mah11:
OK, done. Let me know if I need to do anything else.

dougli1sqrd:
That looks good to me, @kltm?

kltm:
@dougli1sqrd If it looks good to you, go ahead and take it.

PR #1585, opened 2020-11-19
Update MEROPS links

pgaudet:


kltm:
These look okay to me.

balhoff:
Agree!

PR #1592, opened 2020-12-02
allows ONTOLOGY variable to be set from the environment

dougli1sqrd:


kltm:
This is intended to go along with geneontology/pipeline#200

kltm:
Adding in for "immediate" testing.

PR #1594, opened 2020-12-07
Update group-contacts.csv

suzialeksander:
adding user from https://github.com/geneontology/helpdesk/issues/291


kltm:
As with https://github.com/geneontology/go-site/pull/1596, is this still open/ongoing?

PR #1595, opened 2020-12-07
Update users.yaml

suzialeksander:
New user from https://github.com/geneontology/helpdesk/issues/291

User did not indicate an interest in using Noctua

### DIFF
--- a/metadata/users.yaml
+++ b/metadata/users.yaml
@@ -4224,4 +4224,15 @@
   nickname: 'David W. Sant'
   organization: University of Utah
   uri: 'https://orcid.org/0000-0001-7372-9896'
-  
+  -
+  accounts:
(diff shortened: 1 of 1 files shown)


kltm:
@suzialeksander I'm guessing that that groups is not in groups.yaml, or badly formed?

kltm:
@suzialeksander I'm also not sure if groups or authorizations are necessary or desired here, as they do not seem to be sponsored to work with Noctua?

suzialeksander:
Correct, I did not give them edit permissions in Noctua.  It's a new group, I did attempt to add to groups.yaml but I was unsure about that too.

kltm:
Okay, since they are not Noctua, I'm pulling the authorizations and groups.

PR #1596, opened 2020-12-07
Update groups.yaml

suzialeksander:
adding NCBI for https://github.com/geneontology/helpdesk/issues/291, but can someone confirm we need this groups.yaml entry if they aren't asking for Noctua access? This pull is paired with https://github.com/geneontology/go-site/pull/1595 users.yaml update.

### DIFF
--- a/metadata/groups.yaml
+++ b/metadata/groups.yaml
@@ -410,3 +410,7 @@
   label: DiBest-Unical
   id: https://www.unical.it/portale/strutture/dipartimenti_240/dibest
   shorthand: DIBU
+-
+  label: NCBI
+  id: https://www.ncbi.nlm.nih.gov/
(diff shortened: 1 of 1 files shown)


pgaudet:
I dont know what are our rules for groups - @kltm @cmungall can you provide guidance ? 
We should also document this somewhere.

kltm:
@suzialeksander For the other PR to pass, this would probably need to be over there.
You do not need groups or authorization for a non-noctua user.

kltm:
@suzialeksander Is this still open?

suzialeksander:
I'm going to kill this request, it wasn't entirely clear this user is Noctua-ing and if they haven't run into a problem yet they clearly don't need this.  Can always add again later if we have a problem .

PR #1607, opened 2021-01-13
Update db-xrefs.yaml

dosumis:
Created to accurately reflect URL structure for versioned wikipedia links.  These are useful for defs where we want to be able to point to the version used for a def, rather than whatever is current.

### DIFF
--- a/metadata/db-xrefs.yaml
+++ b/metadata/db-xrefs.yaml
@@ -2805,6 +2805,16 @@
       url_syntax: http://en.wikipedia.org/wiki/[example_id]
       example_id: Wikipedia:Endoplasmic_reticulum
       example_url: http://en.wikipedia.org/wiki/Endoplasmic_reticulum
+- database: Wikipedia_versioned
+  name: WikipediaVersion
+  generic_urls:
+    - http://en.wikipedia.org/
+  entity_types:
+    - type_name: entity
+      type_id: BET:0000000
+      url_syntax: https://en.wikipedia.org/w/index.php?title=[example_id]
+      example_id: Wikipedia:Endoplasmic_reticulum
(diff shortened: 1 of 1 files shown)


cmungall:
do you really need CURIEs for these? Can you use the complete URL in the def xref?

I'm not necessarily against it, but it raises a few issues. 

We have a lot of unusual characters in the local_id - we don't explicitly prohibit these but then this is underspecified.

We will now have a choice of class-level xrefs for wikipedia. historically we have used the plain wikipedia URL/xref. Should we switch to versioned for this?

There are other cases where we might want to specify a version of a def xref. PMIDs should be static, but web URLs are used frequently, as well as tracker URLs. ISBNs don't resolve to a static entity (different versions of textbooks, encyclopedias). Maybe we want an orthogonal mechanism for versioning? But even for a PMID, these are often interpreted through a prism of current knowledge.

Also IMO versioned wikipedia is an odd level of specificity. It's already a 3ary source. My own convention is to either follow the citations to get to more direct sources, or just don't let perfection be the enemy of the good and use the wikipedia URL.

Don't want to block important work though, lmk if this is important for you if so we can merge and carry on discussion elsewhere eg obo slack

pgaudet:
Would we need to change the existing Wikipedia links ? Or would we have both versioned and unversioned links?

dosumis:
Would really like to have this.  I think fine to have both versioned & un-versioned wikipedia links, but I have come across cases where versioning would have been useful because of subsequent changes. 

> Don't want to block important work though, lmk if this is important for you if so we can merge and carry on discussion elsewhere eg obo slack

OK.  Will merge.   Happy to discuss more on OBO slack.

PR #1614, opened 2021-02-05
adding 2.2 as gaf output version in gaf reprocess step

dougli1sqrd:
for a partial fix of https://github.com/geneontology/pipeline/issues/212.


kltm:
Looks good.

Issue #1558, opened 2020-09-23
gorule-0000059 Upgrade 2.1 GAF files with GAF 2.2 default formula


kltm:
Proposed action as GORULE:0000059: upgrade 2.1 GAF files with GAF 2.2 default formula

* MF: enables
* BP: acts upstream of or within
* CC
  * CC anatomical structure: located in
  * CC protein-containing complexes: part of 

If an annotation line does not have a gp2t qualifier, the above formula should be used to add the information.

This is specifically for dealing with 2.1--2.2 files without gp2term relation will still be treated as an error.

dougli1sqrd:
This will be implemented in the GAF 2.1 parser, and the rule message will be placed in the report for gorule-0000059. While normally gorule-0000001 is used during parsing, we will do this just to let people know we have upgraded their relations/qualifiers. This will be a warning.

pgaudet:
You will be doing a repair, wont you ? Then it's not a warning ?

kltm:
Noting here that we'll want to look at gafrencer as well once the rules are finalized. This is not an issue in the current flow of the pipeline, but it's good to keep things aligned. @balhoff

kltm:
Noting that this will not technically be "necessary" until the March, but must be implemented and tested before then.

dougli1sqrd:
So just to clarify the above rule for CC:
If the GO Term is subClassOf GO:0110165 "cellular anatomical entity" then we use relation "located in".
If the GO Term is subClassOf GO:0032991 "protein-containing complex" then we use "part of".

In Ontobee, it looks like GO:0019012 "virion", GO:0044217 "other organism part", GO:0044423 "virion part" are also direct children of Cellular Component. What should the relation be if it's not in the above to subClassOf closures?

kltm:
@dougli1sqrd 
I believe a better way of formulating the instruction is:
CC protein-containing complexes and children: part_of
Everything else: located_in

dustine32:
Yeah, that @kltm "is-complex/is-not-complex" logic is pretty much what I do for the PAINT GAF 2.2 files.

dougli1sqrd:
That makes sense

kltm:
It looks like with the GAF 2.2 switchover, most sources now fail our sanity checks due to severe output line reduction. Somebody might want to look a little more deeply into that, but I'm assuming that this is an expression of this ticket?

pgaudet:
I need to make a separate error file format 2.1 to test this

[...]

PR #1616, opened 2021-02-09
adding gorule-0000059 text

dougli1sqrd:
For https://github.com/geneontology/go-site/issues/1558


kltm:
@dougli1sqrd Looks fine for a first pass. To keep things quick, merge PR on pass.

dougli1sqrd:
Hi @pgaudet, can you verify you're okay with the gorule-0000059 text as written here?

This is for https://github.com/geneontology/go-site/pull/1616

This has been merged already because we need it for GAF 2.2. But as far as the specific wording, etc, can you look it over? If there's anything to change we can make a ticket to update the specifics, thanks!

pgaudet:
- Isn't that GAF2.2 ? 
- Maybe instead of "This is due to the other annotation formats having a richer set of Relations available to them. In order to interoperate between GAF 2.1 and other formats as other formats have a richer set of allowed relations, this will be a process to provide basic compatibility." , we can start the text of rule with "The GAF2.2 format requires an annotation relation in the qualifier field (column 4). " (and then continue with "When processing..." 
OK ? 

@vanaukenk 
-

pgaudet:
ALSO 

rule is 'implemented', not 'proposed'

PR #1618, opened 2021-02-10
ontobio gaf 2.2 extensions output fix in 2.4.1

dougli1sqrd:


kltm:
Just a bump. Forcing.

PR #1622, opened 2021-02-19
adding some memory probes to just before and after blazegraph command…

dougli1sqrd:
…s when loading ontologies in the makefile


kltm:
Simple changes to better understand the space environment.
Intended to be reverted (?) once space issue identified and fixed.

PR #1623, opened 2021-02-19
Update Makefile

kltm:
Backtrack a little on https://github.com/geneontology/go-site/pull/1622, removing di


dustine32:
@kltm Yep, straightforward change. Looks OK.

Issue #1624, opened 2021-02-19
Delete goa_uniprot_all.gaf earlier in pipeline to free up space


dustine32:
Delete immediately after the goa_uniprot_all_noiea.gaf file is created from it. This should hopefully allow other processes in the mega-make pipeline step to complete without running out of disk space.

kltm:
Even if it doesn't, we want to prevent the file from existing too long because of https://github.com/geneontology/pipeline/issues/207

kltm:
I wonder about lines like: https://github.com/geneontology/pipeline/blob/624c42671ed719e19a4f923ec243bc6e2bfd6eb6/Jenkinsfile#L566

kltm:
Will see effect in next attempt.

PR #1625, opened 2021-02-19
Rm larger uncompressed goa_uniprot_all right after use - issue #1624

dustine32:
For #1624


kltm:
Noting that this is probably a better fix for https://github.com/geneontology/pipeline/issues/207

PR #1629, opened 2021-02-22
Update users.yaml

suzialeksander:
Editing as per @zebrafishembryo 's request through helpdesk.  I think groups and organization are optional fields, and previous-groups is in the schema although I don't see use of  previous-groups anywhere else in this yaml??

### DIFF
--- a/metadata/users.yaml
+++ b/metadata/users.yaml
@@ -617,8 +617,14 @@
 -
   accounts:
     github: zebrafishembryo
+  authorizations:
+    noctua:
+      go:
+        allow-edit: true
   email-md5:
     - 211b318809e38f23520d0bd64f4a2330
+  groups:
+    - 'http://geneontology.org'
   previous-groups:
     - 'http://www.ebi.ac.uk/GOA'
   nickname: 'Alice Dashow'
--- a/metadata/users.yaml
+++ b/metadata/users.yaml
@@ -617,12 +617,6 @@
 -
   accounts:
     github: zebrafishembryo
-  authorizations:
-    noctua:
-      go:
-        allow-edit: true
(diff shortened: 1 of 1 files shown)


kltm:
@suzialeksander While not in the schema, there is a secondary sanity check (`python3 ./scripts/sanity-check-users-and-groups.py --verbose --users metadata/users.yaml --groups metadata/groups.yaml`) which requires users to have a group. We put that in there for a reason (probably); I'm trying to think why we do that...

kltm:
@suzialeksander Okay, looking at the code a little, I think I know what we're doing and why it's failing. As a general rule, the entries in users.yaml do not require a group, etc., so it is not in the schema. However, as a matter of GO/Noctua policy, it was decided some time ago that groups were required for noctua editors (as opposed to other entries); as part of enforcing that policy, this script was added.
If we want to revisit that, we should probably do that formally--that assumption may be bolted in other places now.

kltm:
I'd also note that something in "groups" can be quite informal and trivial, not necessarily something as large as a MOD or other resource.

suzialeksander:
@vanaukenk can we get an "official" Noctua approval for this change?

suzialeksander:
I've added GOC as Alice's `group`, as I think we discussed.  Can @kltm and @vanaukenk verify this is ok, and ok to merge?

vanaukenk:
@suzialeksander - this is a question for @thomaspd 
Let's confirm with him first (unless you've already heard from him).

suzialeksander:
Sounds like GOC is the solution for now. @kltm ok to merge?

kltm:
Sounds good. Discussed with @vanaukenk and @pgaudet

PR #1631, opened 2021-02-22
Pipeline 206 scripts

dougli1sqrd:


kltm:
re: geneontology/pipeline#206

dougli1sqrd:
Okay this is how I would like it now

kltm:
Taking in cuz it looks safe.

PR #1642, opened 2021-03-01
update mgi.yaml gpad source url to correct file name

dougli1sqrd:
Just a very slight filename correction in the MGI gpad source URL. Previously named `gpa`, now named `gpad`.


kltm:
Correct reportedly works.

PR #1646, opened 2021-03-04
update to ontobio 2.6.0 with gorule-0000061 implemented

dougli1sqrd:


kltm:
Taking for gorule-0000061

PR #1647, opened 2021-03-04
Update zfin.yaml

sabrinatoro:
@kltm 
I updated the source for the ZFIN GAF. This GAF is the 2.2 format.


kltm:
@dougli1sqrd Noting the ...`2.2.gaf.gz` name. If that's not problematic, please go ahead and merge.

kltm:
For https://github.com/geneontology/pipeline/issues/217

dougli1sqrd:
Yeah this is fine! We rename once we download anyway.

PR #1648, opened 2021-03-05
Update gorule-0000061.md

vanaukenk:
Fixing id for root molecular_function (was citing alt_id)


vanaukenk:
@dougli1sqrd @pgaudet 
I was trying to just fix a typo in gorule-0000061, but it looks like the error messages I'm getting have to do with some of the tests for other rules.
Do we need to address those errors before I can merge this fix?

kltm:
@dougli1sqrd @vanaukenk Forcing text change merge

PR #1650, opened 2021-03-09
Update gorule-0000061.md

dougli1sqrd:
Updating cellular component GO ID to `GO:0005575` rather than deprecated GO:0008372.


dougli1sqrd:
I will also point out here that the rule states that *any* term within Cellular Component is repaired and *never* filtered, and in Molecular Function or Biological Process we only repair if the term is a root, but filter if not.

So this rule will cause annotations only in MF and BP to be possibly filtered, and never in CC, which are always repaired. Was this intentional?

kltm:
Textual change, forcing merge.

pgaudet:
> I will also point out here that the rule states that any term within Cellular Component is repaired and never filtered

Maybe we did not formulate the rule correctly, but we wanted at least some CC to be filtered:
> The gp2term relation RO_0002325 "colocalizes_with" for "GO:0032991 "protein-containing complex" is not allowed and is filtered.

Interestingly, I dont see where we asked for filters in BP and MF. 
Let's discuss this today.

PR #1651, opened 2021-03-09
update ontobio to 2.6.1 for gorule 61 bugfix

dougli1sqrd:
@kltm update version, for the history books


dougli1sqrd:
Use the force, Luke, and force merge

PR #1652, opened 2021-03-09
Update gorule-0000001.md

pgaudet:
To refer to GAF2.2

### DIFF
--- a/metadata/rules/README.md
+++ b/metadata/rules/README.md
@@ -8,7 +8,7 @@ contains the information required to execute it.
 For more details for GOC members on how to create rules, see [SOP.md](SOP.md)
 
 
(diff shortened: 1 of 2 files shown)


dougli1sqrd:
Should we generalize this for GPAD as well? Or like, to say that whatever annotations we're parsing should be parsable according to the available spec? I see this as generalizing across all our data formats.

pgaudet:
> Should we generalize this for GPAD as well?

I suppose this is not a question for me? As I mentioned before I dont like gorule-0000001 because it's too vast, and the vague description (everything is parsed according to specs) is not quite true. 
I propose to have a separate general GPAD rule, so that the error reports are clear about the errors found. 
Alternatively we can add a parameter somewhere in the header that states which file type was parsed.

PR #1657, opened 2021-03-09
Update gorule-0000013.md - add implementation note on how taxon constraints are checked

cmungall:


pgaudet (commented):
> The gaferences files is processed in the pipeline via ontobio and includes alongside other GO rules reports.

should be  **"included"** ?

dougli1sqrd:
I formatted the gaferencer json example as well as added an `implementations` entry in the metadata portion of the yamldown

Issue #1649, opened 2021-03-09
gorule-00000008 example is failing, but the test example is actually correct


dougli1sqrd:
One of the gorule-00000008 examples is failing.

`UniProtKB	Q9WVH3	Foxo4		GO:0031572	PMID:12048180	IDA		P	Forkhead box protein O4	Foxo4|Afx|Afx1	protein	taxon:10090	20110425	MGI` is expected to fail, but is passing in the example test run in travis. 

But when I run it locally, gorule-00000008 is correctly triggered, giving the message: `"message": "GORULE:0000008: No annotations should be made to uninformatively high level terms"`. This is expected. 

Offending example is https://github.com/geneontology/go-site/blame/master/metadata/rules/gorule-0000008.md#L19

kltm:
@pgaudet The issue for the failure here was that the given term for the test annotation line (https://github.com/geneontology/go-site/pull/1662/files#diff-5830fda4cd89ac90f5ef8cba844b5b14772472046e21502313b6ffc0cc6247a2R19) has been replaced (by a term that is no longer in that subset). I selected a new term from that subset, as we wanted PRs to work again, but you might prefer a real example or better term?

Issue #1660, opened 2021-03-10
Travis tests are currently failing on final test


kltm:
Currently, the travis tests seem to be pretty consistently failing on:

`docker run -it -v $(pwd):/tmp/go-site -w /tmp/go-site geneontology/dev-base:latest /bin/bash -c "pip3 install -r pipeline/requirements.txt && ./scripts/rule-example-validation.sh"`

Testing locally, I also get this failure.

Issue may be:

 ```json
        {
            "rule": "gorule-0000008",
            "type": "fail",
            "format": 2,
            "input": "UniProtKB\tQ9WVH3\tFoxo4\t\tGO:0031572\tPMID:12048180\tIDA\t\tP\tForkhead box protein O4\tFoxo4|Afx|Afx1\tprotein\ttaxon:10090\t20110425\tMGI",
            "expected": false,
            "actual": true,
            "success": false,
            "reason": "Repair found `True`, but expected `False`"
        }
```

Tagging @dougli1sqrd

dougli1sqrd:
Yeah, I documented this here as well: https://github.com/geneontology/go-site/issues/1649

kltm:
Cleared.

PR #1662, opened 2021-03-11
Update gorule-0000008.md

kltm:
Correct comment and update second example with GO:0000910, which actually is in the subset. GO:0031572 has been deprecated and the replacement term is not in the subset.
Tagging @pgaudet in case there is a better term to use here--I just picked one randomly that satisfied the criteria.


kltm:
This will fix #1649 and #1660

dougli1sqrd (approved):

PR #1666, opened 2021-03-22
renaming noctua gpad output to not include 'valid' and use -src

dougli1sqrd:
This is for the name update of noctua gpad files. Output is now `noctua_[dataset].gpad.gz`. Inputs should be named `noctua_[dataset]-src.gpad.gz`.

https://github.com/geneontology/pipeline/issues/220


dustine32 (approved):
Thanks @dougli1sqrd! This looks like it should fix the names.

kltm:
Seems good.

PR #1667, opened 2021-03-23
fixes an error in how we processed file names in the makefile for the…

dougli1sqrd:
… noctua post processing

For https://github.com/geneontology/pipeline/issues/220


kltm:
Give it a whack.

Issue #1668, opened 2021-03-25
Add a flag option to control sanity checks on Noctua annotation files


dustine32:
Due to the current situation where we produce a `noctua-*-src.gpad` file but we don't want to run sanity checks on them using script `sanity-check-ann-report.py`, we should add a flag option that skips any file containing "noctua" in its basename.

PR #1669, opened 2021-03-25
Ignore noctua annotation files when running sanity checks

dustine32:
For #1668


kltm:
@dustine32 assures me.

PR #1671, opened 2021-04-01
Update goref-0000116.md

cmungall:
Always use standard URLs.. maybe should use PURLs?


kltm:
@cmungall I'm curious about this. I don't think I would have considered http://geneontology.org/external2go/ontology/external2go/rhea2go to be a purl really--just a file on a site that we've passed forward through time. http://current.geneontology.org/ontology/external2go/rhea2go seems like a more logical place for the file, in with the rest of them and as part of a release package. Naturally, things like current and snapshot are not really purls either, but they are where we generally send people (and likely what our purls would generally resolve to).
It might be good to have a list of the resources that we have that are regarded as purls, so we (I) don't confuse them with "just" files, etc.? Also, do you think it might also be worth considering either rolling our own (purl.geneontology.org) for important things or using OBO? I think there can sometimes be confusion (with migrations, redirects, maintenance) when critical and normal paths get mixed.

Issue #1590, opened 2020-12-02
Switch go-site repo from travis to github actions


kltm:
Travis has left us in a lurch. As an immediate short to medium term solution, we'll be switching to GH actions for go-site checks.

kltm:
Tagging @dougli1sqrd @dustine32 for taking a look at the PR before Travis turns off.

PR #1672, opened 2021-04-01
Pipeline issue 206 downloader update

dougli1sqrd:
smart downloading!

new command called `annotations` that uses either gaf, or gpad+gpi.

for https://github.com/geneontology/pipeline/issues/206

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -16,6 +16,19 @@ Additionally, this will only download "active" datasets - datasets whose YAML ha
 
 As a note, the `all` command differs in that it will download all files of a certain type, and so will either repeat data by downloading both gpad and gaf (if both are provided)
 
+#### Examples
+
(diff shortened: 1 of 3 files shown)


kltm:
@dougli1sqrd It's looking good.

Keeping in the spirit of "locally executable atomic parts running in order", would you mind adding a README with a few examples of usage and the expected inputs and outputs?

dougli1sqrd:
Sure. Where should the readme live? Or which readme should I add to?

dougli1sqrd:
I also have concerns about the name of the script referring to GAF still, rather than something more generic. But if we renamed it we'd have to change the invocations too

kltm:
Changing invocations I think is fine to keep things clear. We're ripping up the floorboards here anyways, so why not?
For a README, either in the main one in scripts, or README.script_name.

dougli1sqrd:
okay @kltm, added some examples for downloader script readme

kltm:
Given that this is hung up on docker/travis issues, please go ahead and merge. Also see https://github.com/geneontology/go-site/issues/1590

PR #1675, opened 2021-04-14
updating ontobio to 2.7.0

dougli1sqrd:
Main change here is ability to specify "rule sets". Be default, the main pipeline kernel will still behave the same with no change. Other scripts that perform annotation parsing (like ontobio-parse-assocs) will be default not run rules.

### DIFF
--- a/pipeline/requirements.txt
+++ b/pipeline/requirements.txt
@@ -1,3 +1,3 @@
 PyYAML
-ontobio==2.6.3
+ontobio==2.7.0
 click


kltm:
@dougli1sqrd If you think this is safe to merge just before we run a `release`, go for it.

dougli1sqrd:
I don't think there's much risk, but let's just be cautious and wait until after the release.

sierra-moxon:
closing as ontobio is up to 2.8.0 in master

Issue #1530, opened 2020-07-02
Reinstate go-ontology and go-annotation summary emails


kltm:
We are going to reinstate the go-ontology and go-annotation summary emails, which had been left behind when the old pipeline system finally went offline. The jobs were: go-github-annotation-tracker-summary-email and go-github-ontology-requests-summary-email.

For this, we'd like to keep it out of the main pipeline, as they are solely a function of GitHub, and are looking to implement them with GH actions. As well, we'd like to take this chance to migrate from the crufty old JS to standard and no-frills python.

Originally https://github.com/geneontology/helpdesk/issues/259

Tagging @dustine32 @dougli1sqrd

kltm:
Operationally:
### go-annotation
```
Cron:
      <spec>H H(20-23) * * *</spec>

Commands:
npm install
node ./query-github-for-new-tickets.js -p go-annotation > ../../body.txt

Email profile:
          <email>
            <recipientList>go-consortium@lists.stanford.edu,go-curator-tracker@l
ists.stanford.edu</recipientList>
            <subject>Github Annotation Tracker Update</subject>
            <body>${FILE,path=&quot;body.txt&quot;}</body>
            <recipientProviders/>
            <attachmentsPattern></attachmentsPattern>
            <attachBuildLog>false</attachBuildLog>
            <compressBuildLog>false</compressBuildLog>
            <replyTo>$PROJECT_DEFAULT_REPLYTO</replyTo>
            <contentType>project</contentType>
          </email>
```
### go-ontology
```
Cron:
      <spec>H H(20-23) * * *</spec>

Commands:
npm install
node ./query-github-for-new-tickets.js > ../../body.txt

Email profile:
          <email>
            <recipientList>go-consortium@lists.stanford.edu,go-curator-tracker@lists.stanford.edu</recipientList>
            <subject>Github Ontology Tracker Update</subject>
            <body>${FILE,path=&quot;body.txt&quot;}</body>
            <recipientProviders/>
            <attachmentsPattern></attachmentsPattern>
            <attachBuildLog>false</attachBuildLog>
            <compressBuildLog>false</compressBuildLog>
            <replyTo>$PROJECT_DEFAULT_REPLYTO</replyTo>
            <contentType>project</contentType>
          </email>
```

kltm:
See https://github.com/geneontology/go-site/blob/master/scripts/jenkins/query-github-for-new-tickets.js

pgaudet:
[...]

PR #1679, opened 2021-05-05
GH issue search now full 24hr, for #1530

dustine32:
The GH query is time-sensitive now. So, running this at 4PM Pacific on 2021-05-05 to look back one (1) day will retrieve issues created or updated between 2021-05-04 4PM Pacific and 2021-05-05 4PM Pacific.


kltm:
The little rascal works for me too.

Issue #1655, opened 2021-03-09
clarify gorule-0000023, annotation propagation, regarding propagation of redundant annotations


cmungall:
https://github.com/geneontology/go-site/blob/master/metadata/rules/gorule-0000023.md

includes:

```
TBD
Should this pipeline filter annotations based on some redundancy criteria?
```

(I think the intention is to say "Should this _rule_ filter"...)

I don't think any rule should have a TBD in it. If it is an active rule, everything should be decided. However, we can record past decisions, for example:

HISTORY

There was previous discussions as to whether this rule should do X or Y, see ticket <URL>

In this case we should make the rule very precise

A complication here is that the current *implemented* behavior is to filter redundant predictions. See https://github.com/geneontology/go-ontology/issues/19954

pgaudet:
I have created a PR with the following modifications: 
1. Rule = implemented, rather than proposed
2. Update rule justification ("background")
3. Update Procedure and Provenance sections. 

TO DO: 
Implement as follows: 
 - [ ] Any asserted or inferred annotation to a **MF**, where MF part-of BP, will generate an `involved-in` to that BP
 - [ ]  Asserted or inferred annotation to a **BP using the `involved-in` relation**, where BP occurs-in CC, will generate a `part-of` annotation to that CC
 - [ ] ECO (GAF2.2 Column 7/GAPD1.1 Column 6): IEA
 - [ ]  "with/from" (GAF2.2 Column 8/GAPD1.1 Column 7): the GO term from which the anontation is propagated
 - [ ]  Reference: GO_REF:0000108 (https://github.com/geneontology/go-site/blob/master/metadata/gorefs/goref-0000108.md)
 - [ ]  Date (GAF2.2 Column 14/GAPD1.1 Column 9): the date the script was run
 - [ ]  Assigned_by (GAF2.2 Column 15/GAPD1.1 Column 10): GOC

Thanks, Pascale

vanaukenk:
@pgaudet 

For the BP 'occurs in' CC, can you clarify?  Do we instead want to use the gp2term relation 'located in' or maybe even 'is active in'?

Do we need to look at some examples to see if this always holds?

pgaudet:
You're right, I should have stated 'is active in'. I'll edit the specs.

pgaudet:
As a reference - noting that GOA is predicting annotations from interontology links for the following relations: 
- part_of and occurs_in

pgaudet:
[...]

PR #1683, opened 2021-05-10
Update gorule-0000023.md

pgaudet:
@dustine32 @kltm 
We should implement those fixes - this is an Action Item from the May 2021 GOC meeting.


kltm:
@pgaudet To clarify, is this a work request as PR or is there a ticket as well?

pgaudet:
#1655

PR #1685, opened 2021-05-18
Add "added/removed" PMIDs to go-stats-summary for visualization on GO website

lpalbou:
Add the "added/removed" PMIDs and references to go-stats-summary, and therefore in new aggregated-go-stats-summaries

https://github.com/geneontology/go-stats/issues/15


lpalbou:
@kltm you may want to test run that on snapshot. I updated the networkx call from .node -> .nodes as it was deprecated in networkx 2.4 and .nodes was here with version 2.0 so it should be fine. If not, I will revert back to .node but we should keep in mind that whenever the pipeline update the version of networkx, we have that deprecation of a very common method.

kltm:
@lpalbou Testing on https://build.geneontology.org/job/geneontology/job/pipeline/job/master/632/

kltm:
Seems to work. Merging.

Issue #1697, opened 2021-06-03
Incorrect interontology linksbased on NOT annotations


pgaudet:
Hello, 

This was reported in the annotation tracker: 
https://github.com/geneontology/go-annotation/issues/3799

> UniProtKB:Q1321 ERCC8
NOT|enables GO:0003678 DNA helicase activity
ECO:0000314 IDA PMID:8999876 9606 Homo sapiens UniProt

> GO is using NOT annotation when creating annotation via F-P links
Which gives:
UniProtKB:Q13216 ERCC8 involved_in GO:0032508 DNA duplex unwinding ECO:0000364 IEA GO_REF:0000108
GO:0003678 9606 Homo sapiens GOC

I updated the rule to reflect this, but we need to code that modification
https://github.com/geneontology/go-site/blob/master/metadata/rules/gorule-0000023.md

Thanks, Pascale

kltm:
https://github.com/geneontology/go-site/pull/1698

kltm:
Looks like this is still predictions in owltools.
Options:
1) try and fix in owltools
2) drop owltools and switch to ontobio-based predictions

https://github.com/geneontology/pipeline/issues/221
https://github.com/geneontology/pipeline/issues/130

kltm:
@pgaudet Clarifying action here in light of the addition to the GO rule, a fix could be just having "NOT" annotations filtered out during the process?

kltm:
Naively, I would expect this to work:
https://github.com/owlcollab/owltools/blob/05524b3c6b5effeddffa8d0c5f53b8e537ff7a33/OWLTools-Annotation/src/main/java/owltools/gaf/inference/BasicAnnotationPropagator.java#L665 ...

@pgaudet Where is this data coming from? What file? We do not seem to have any such entity in our predictions in the last release:
```
skyhook@wok:~/release/products/annotations$ cat  *-prediction.gaf | grep -c "Q1321"
0
```

pgaudet:
@kltm the Accession number was truncated - it's Q13216


Still I cannot find this either. @ValWood where did you see this?

ValWood:
https://www.ebi.ac.uk/QuickGO/annotations?geneProductId=Q13216&aspect=biological_process


[...]

PR #1698, opened 2021-06-03
Update gorule-0000023.md

pgaudet:


kltm:
https://github.com/geneontology/go-site/issues/1697

PR #1700, opened 2021-06-09
Update users.yaml

hdrabkin:
Adding new user Cailey Mitchell for MGI

### DIFF
--- a/metadata/users.yaml
+++ b/metadata/users.yaml
@@ -4303,3 +4303,23 @@
   nickname: 'Yves Snijckers'
   organization: Radboud University
   uri: 'https://orcid.org/0000-0001-6913-5104'
+-
+  accounts:
+    github: mitchellcailey
+  authorizations:
+    noctua:
+      go:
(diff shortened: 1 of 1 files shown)


kltm:
@hdrabkin I'd note that the `email-md5` should not be the user's email address, but the md5sum of that address, like `a024b57ae611de92ae83ba17ac1a9b4a`. You should do ahead and delete that stanza.

As well, I'm not sure if we're still minting new GO xrefs (i.e. `'GOC:cmm'`) or if it's needed? Any thoughts @vanaukenk ?

vanaukenk:
@kltm @hdrabkin 
At this point, I wouldn't mint a new GO xref (i.e. GOC:cmm).  Certainly for Noctua, we only need the orcid and github handle.

hdrabkin:
For the GOC ref I was thinking of eventual import into our interface, the orc id goes into a stanza, but the GOC id is displayed to easily find . 
So should I re-edit the entry?

PR #1720, opened 2021-07-26
Update db-xrefs.yaml

mah11:
add entry for AlphaFold at EBI


mah11:
I copied the ID syntax from the UniProtKB entry, because the links to individual entries use UniProtKB accessions.

PR #1725, opened 2021-07-30
add JaponicusDB ID & URL details

mah11:


metadata/db-xrefs.yaml
```diff
@@ -2932,3 +2932,14 @@
...
       example_id: AlphaFold:Q5VSL9
       example_url: https://alphafold.ebi.ac.uk/entry/Q5VSL9
+- database: JaponicusDB
+  name: JaponicusDB
+  generic_urls:
+    - http://www.japonicusdb.org/
+  entity_types:
+    - type_name: gene
+      type_id: SO:0000704
+      id_syntax: SJAG_[0-9]{5}
```
kimrutherford, line 2942: > id_syntax: SJAG_[0-9]{5}

To also handle the mitochondrial genome we'd need something like:

```
SJAG_[0-9]{5}|SJAG_MIT[0-9][0-9]
```

We might also need to consider the tRNA identifiers from: https://github.com/japonicusdb/japonicus-config/issues/39

mah11: Good point. Is there a decent regex for the tRNA IDs? If so, I'll add it along with the mito version - I'll edit the file and update this PR.

mah11: p.s. we don't have to get the regexes resolved now -- the `id_syntax` line appears to be optional, so we could leave it out for now and add it back after we get ncRNA IDs sorted

kimrutherford: > Is there a decent regex for the tRNA IDs? 

Not yet I think.

> p.s. we don't have to get the regexes resolved now -- the id_syntax line appears to be optional, so we could leave it out for now and add it back after we get ncRNA IDs sorted

OK.  It might be best to do that.  We can add it back once everything's a bit more stable.

mah11: >>  The `id_syntax line` appears to be optional, so we could leave it out for now and add it back ...
>
> OK. It might be best to do that. 

That's done now.

mah11:
> there does not seem to be a typical `namespace:local` setup ...

Sorry, @kltm, I don't know what that means. In principle we're happy to make any necessary adjustments to the JaponicusDB metadata entries, but I can't until I understand what you're asking for.

PR #1726, opened 2021-08-23
Update sgd.yaml

suzialeksander:
Trying to update from no GPI to our in-house GPI 1.2


kltm:
For https://github.com/geneontology/noctua/issues/729

kltm:
Reality check to future self and @dustine32 . I believe that adding the GPI like this does /nothing/ to the regular pipeline, but may make it available to the NEO pipeline with ./datasets.json later on. Adding it now to try an get it through with pipeline runs so we can get this into NEO.

PR #1729, opened 2021-08-24
Ensure model deletion query is not sensitive to string datatype

balhoff:
See https://github.com/geneontology/noctua/issues/736

### DIFF
--- a/pipeline/sparql/delete/delete_non_production.sparql
+++ b/pipeline/sparql/delete/delete_non_production.sparql
@@ -24,7 +24,8 @@ WHERE {
       # Find production graphs
       GRAPH ?g {
         ?g :graphType :noctuaCam .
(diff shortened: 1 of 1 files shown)


balhoff:
@kltm can we test this in a pipeline run?

kltm:
@balhoff Depending on stomach for risk, I might just add it to `master` and let it run through naturally. We could even kill `snapshot` right now and restart it immediately.

balhoff:
I feel confident, go with either of those.

kltm:
@balhoff Okay, I'm going to 1) merge, 2) stop current `snapshot`, and 3) immediately start a new `snapshot` so we can see full products (which would maybe be available starting this time tomorrow or a little after, depending if anything else goes "boink").

PR #1731, opened 2021-08-31
Add .gz suffix to PomBase GPAD and GPI files in pombase.yaml

kimrutherford:


kimrutherford:
See https://github.com/geneontology/noctua/issues/727#issuecomment-908775024

kltm:
Adding `compression` values.

kltm:
Great--thank you!

PR #1735, opened 2021-09-17
Updated id syntax on QTLdb and CorrDB

zhilianghu:
Removed text part: e.g. QTL_ID:23155 >> 23155; Corr_ID:37232 >> 37232


pgaudet:
@kltm FYI

PR #1767, opened 2022-02-01
FIxed GeneDB to PlasmoDB

pgaudet:
FIxed GeneDB to PlasmoDB for Plasmodium falciparum 3D7

Issue #1768, opened 2022-02-01
Datasources: Fix GeneDB data sources


pgaudet:
Hi, 

GeneDB has been taken down, Plasmodium falciparum 3D7 data is now on PlasmoDB. 

I have update the db-xrefs.yaml to reflect that

@kltm @dustine32 @sierra-moxon Does anything need to be updated in the pipeline or AmiGO? 

Thanks, Pascale

pgaudet:
https://github.com/geneontology/go-site/pull/1767/files

kltm:
Like all of GeneDB is gone? We are still loading some file: https://github.com/geneontology/pipeline/blob/snapshot/Jenkinsfile#L134-L135
Are you saying that these need to be removed?

kltm:
Talked to @pgaudet : will revert changes made to links; no data updates are necessary for data load.

kltm:
Reverting with https://github.com/geneontology/go-site/pull/1770

pgaudet:
In fact this was correct. The links in AmiGO (and QuickGO) to GeneDB should be changed to PlasmoDB.

pgaudet:
@Achchuthan Do you generate a mappings between the genedb and plasmodb identifiers? This is for GOA (QuickGO) to be able to link out. @alexsign has an older version of the file, but if this is maintained and updated, we'd like to be able to get it from you. 

Thanks, Pascale

pgaudet:
I'll open a new ticket for this

kltm:
Dupe to https://github.com/geneontology/go-site/issues/1777

PR #1770, opened 2022-02-01
Revert "FIxed GeneDB to PlasmoDB"

kltm:
Reverts geneontology/go-site#1767


kltm:
For https://github.com/geneontology/go-site/issues/1768

PR #1771, opened 2022-02-02
Added Kalpana from IntAct

pgaudet:


kltm:
Noting that IntAct and InTact both appear in the file.

pgaudet:
Thanks, fixed

PR #1776, opened 2022-02-04
Added Plasmodium back to geneDB dataset

pgaudet:


kltm:
@pgaudet Could you make a ticket out of this and add it to the pipeline project? Currently, all species additions and removals need to be coordinated at different locations.

PR #1787, opened 2022-02-10
Update db-xrefs.yaml

pgaudet:
change https://registry.identifiers.org/registry -> 	https://identifiers.org


kltm:
Created https://github.com/geneontology/go-site/issues/1788

PR #1802, opened 2022-02-24
Update goref-0000117.md

suzialeksander:
typo & clarifying what ARBA is


suzialeksander:
also, reading through this description, I'm a little confused- @alexsign are the fields like DE and LO internal to Uniprot? We're trying to add this GO REF to SGD and I need to make sure we have our text description correct, thanks

pgaudet:
I thought the GO didn't have an ftp site anymore (see last line in the text)

suzialeksander:
good catch.  I've altered that, although we can also just say 'available on the GO and GOA sites' leaving out mention of FTP

PR #1809, opened 2022-03-03
flatten consolidated generated gorefs.yaml

sujaypatil96:
Modify parser to generate a flatter consolidated gorefs.yaml file.


kltm:
Updates to parser; should have no outside effects.

PR #1810, opened 2022-03-04
Create GORULE:0000062 Infer annotations on molecular function via has…

pgaudet:
…_part


kltm:
For https://github.com/geneontology/pipeline/issues/130

PR #1814, opened 2022-03-15
Add `gorefs.yaml` and corresponding JSON Schema validator

sujaypatil96:
This PR seeks to do the following:

- [x] Add consolidated `gorefs.yaml` file created by using the parser written in PR #1785 
- [x] Add JSON Schema that can validate the above YAML file
  - [x] Add appropriate descriptions for each of the properties in JSON Schema
  - [x] Validate multiple document structure in YAML file
  - [x] Add further constraints like regex for validation
- [x] Add LinkML schema 
  - [x] Generate Python artifact
  - [x] Generate JSON Schema artifact

CC: @kltm

PR #1815, opened 2022-03-17
`gorefs.json` similar to `gorefs.yaml`

sujaypatil96:
Modify the goref parser so that it can generate a `gorefs.json` equivalent to `gorefs.yaml` so that it can be validated by the JSON Schema in PR #1814.

### DIFF
--- a/metadata/gorefs/gorefs.json
+++ b/metadata/gorefs/gorefs.json
@@ -1,0 +1,1272 @@
+[
+    {
+        "authors": "GOA-UniProt curators",
+        "id": "GO_REF:0000029",
+        "is_obsolete": true,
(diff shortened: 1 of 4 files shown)


kltm:
@sujaypatil96 You don't necessarily need to worry about PRs into your own namespace.
I'm not quite following how this json is specifically working in our workflow here, but as long as were doing:
(gorefs.yaml, source of truth) -> (checking gorefs.yaml format and fields for correctness) -> (splitting into md) -> (injecting md into go website)
Each of those should probably be a separate runnable command.
If this is for the earlier (one time) step of:
(md) -> (gorefs.yaml)
Whatever works is great :)

PR #1817, opened 2022-03-29
Added DisProt

pgaudet:


kltm:
For https://github.com/geneontology/go-site/issues/1818

PR #1835, opened 2022-04-18
Fix typo in ORCID

balhoff:
The ORCID for Stan Laulederkind has a duplicated segment.


kltm:
@balhoff Happy take whenever you want, but we might want to have a coordinated data update if this has leaked other places?

balhoff:
Does he edit in Noctua? I tried grepping for this ORCID in the models and didn't find it. Where else should we look?

kltm:
@balhoff I believe that should be sufficient.

PR #1836, opened 2022-04-18
Add some missing URIs.

balhoff:
Added several missing ORCIDs, and one Wikidata URI where an ORCID could not be found. This relates to https://github.com/geneontology/go-ontology/issues/21163 and https://github.com/geneontology/go-ontology/issues/22551.

### DIFF
--- a/metadata/users.yaml
+++ b/metadata/users.yaml
@@ -439,7 +439,7 @@
 -
   nickname: 'Michael Ashburner'
   organization: FlyBase
-  uri: 'GOC:ma'
+  uri: 'https://orcid.org/0000-0002-6962-2807'
   xref: 'GOC:ma'
 -
   accounts:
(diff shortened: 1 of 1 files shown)


kltm (changes requested):
As a matter of current policy, we do not allow URIs that are not `GOC` or ORCID.

balhoff:
> As a matter of current policy, we do not allow URIs that are not `GOC` or ORCID.

@kltm okay I changed Amelia back to `GOC`.

kltm:
@balhoff confirming that you think the additon/change is safe with `grep 'GOC:\(lh\|ef\|elh\|kad\|clt\|jib\|ceb\|mb\|ma\)' *.ttl` returning nothing.

balhoff:
Thanks for checking that. These changes will result in changes to http://snapshot.geneontology.org/metadata/users.ttl (which was my goal). Basically `<http://purl.obolibrary.org/obo/go/users/ma>` will become `<https://orcid.org/0000-0002-6962-2807>`. I did this grep: `grep 'http://purl.obolibrary.org/obo/go/users/' *.ttl` in `noctua-modesl/models`. It doesn't seem like we have any of those user IRIs in the models. The end goal is to use these ORCIDs in attributions in the ontology, although I'm running into OBO format irritations there...

kltm:
@balhoff I think this seems okay for me, as it would only affect things outside of our standing data store.

kltm (approved):
Reversion in alignment with current policy.

PR #1837, opened 2022-04-20
suppress owltools summaries

kltm:
for geneontology/pipeline#281


kltm:
Talked to @pgaudet . Will move forward.

PR #1846, opened 2022-04-27
Update requirements.txt ontobio to remove aspect check for GPAD

dustine32:
For https://github.com/geneontology/pipeline/issues/280.


kltm:
Update and go!

PR #1861, opened 2022-05-16
Update group-contacts.csv

suzialeksander:
@vanaukenk and/or @pgaudet - is this group ok to add? Adding for https://github.com/geneontology/helpdesk/issues/392 but just checking that this is the best way to add this new user/group - if this group stops being active is it simple to switch the ownership of annotations back to pombase?


suzialeksander:
I think adding a group like "University of Florida" is more in line with the other groups in this file?

Issue #1877, opened 2022-07-06
We should standardize on a single email address for documentation


kltm:
In the GO documentation, we should standardize on a single email address (domain).

The domain should always be listed as lists.stanford.edu.

kltm:
Also standardizing on the new go helpdesk email address from the gohelp list one.

PR #1878, opened 2022-07-06
Update mail lists to new standard

kltm:
For issue #1877 update mail lists

Tagging @pgaudet @suzialeksander


suzialeksander:
Confirming that both lists.stanford.edu and mailman.stanford.edu are valid, but lists.stanford.edu is preferred.  Looks right.

PR #1890, opened 2022-08-01
Update db-xrefs.yaml for new Xenbase URIs

malcolmfisher103:
This updates the URL syntax for Xenbase to our new simplified approach, the old ones may be non-functional due to rearrangements of our site code.

### DIFF
--- a/metadata/db-xrefs.yaml
+++ b/metadata/db-xrefs.yaml
@@ -2843,27 +2843,27 @@
     - type_name: gene
       type_id: SO:0000704
       id_syntax: XB-(GENE|LINE|TRANSGENE|MORPHOLINO)-[0-9]+
-      url_syntax: http://www.xenbase.org/common/xsearch.do?exactSearch=true&searchIn=7&searchValue=[example_id]
+      url_syntax: http://www.xenbase.org/entry/[example_id]
       example_id: Xenbase:XB-GENE-6255962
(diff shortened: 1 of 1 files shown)


kltm:
http://www.xenbase.org/entry/XB-LINE-1248 goes to a 404 currently for me.

malcolmfisher103:
We are curently in the middle of updating the Xenbase codebase to merge previously distinct modules into one common module. This is an ongoing process and the stock center related entities, lines and strains, module merge has not yet been deployed.

As things currently stand using the old syntax the gene links are not working. I updated all of the link syntaxes as the new syntax should work after our next deployment, as it currently does on our test site (http://test.xenbase.org/entry/XB-LINE-1248). 

Is it better to do this as 2 different pulls a couple of weeks apart? My thinking was that it was more important that the genes and to a lesser exent reagents that we frequently reference had working links over the less frequently referenced lines and strains being non-functional for a week or so until the next deployment, but it can be done piecemeal.

kltm:
@malcolmfisher103 Fair enough--I see now that the gene and morpholino do indeed work (with the other two "still in progress").
I'm happy to stage things however you like in the metadata; as only a single form would be accepted anyways, if this works for you I'm happy to take it now as-is.

malcolmfisher103:
@kltm  Yes. please use this as-is, hopefully everything will be working in the coming weeks.

kltm:
@malcolmfisher103 Understood--merging!

Issue #1891, opened 2022-08-16
Jenkins email issues; ontology summary email missed three days


kltm:
Due it a hardware issue, the ontology summary emails missed three days: August 12, 13, 14. <strike>The emails were sent out as usual on the 15th and the problem has been cleared.</strike>

Tagging @vanaukenk @pgaudet

kltm:
@vanaukenk @pgaudet I'm not sure how much of a disruption this actually is. We can't currently produce a custom email for an arbitrary interval, but it may be something we could pursue if deemed necessary.

kltm:
We may be having this issue currently: https://issues.jenkins.io/browse/JENKINS-69170
This means that pipelines can run, but that they cannot send emails...which will mean that things can run more-or-less as usually if we watch them, but automated summary emails may be off the table for a moment while we figure something else out (tagging @dustine32 ).

pgaudet:
Hi @kltm 

I dont think this is a huge problem, especially it was mostly on the weekend. We can send an email to go-consortium to notify that 3 mails were skipped. 

Thanks, Pascale

kltm:
To help us with this @dustine32 is taking a look at adding an offset to the script.
I think there are practically a few things we can do at this point:
1. send a catchup email (or emails) and an explanation
2. switch over to a manual process, but reduce the frequency, until we're back to normal (I have hopes that the September Jenkins release would clear this)
3. try and get a code fix in to bypass the email system and try and set something else inside the pipelines

kltm:
@dustine32 Has completed a modified script that will allow us to generate the summary for an arbitrary period. We can use this to help "catch up", etc.

pgaudet:
Great!

kltm:
At request from @cmungall , adding a little more to the script to try and put the number of issues at the top. Unfortunately, Jenkins doesn't currently inspect the text or return of the summary script, so I added a flag to make it produce just a useful "title". As Jenkins is borked for the moment, we'll have to test that once (if) things come back. Otherwise, we can use it to produce the interim emails.

kltm:
Email list: go-consortium lists stanford edu.
Get title for go-ontology:
`reset && python3.6 ./scripts/github_issue_summary.py geneontology/go-ontology 7 --summary_only true`
Get email contents for go-ontology:
`reset && python3.6 ./scripts/github_issue_summary.py geneontology/go-ontology 7 > /tmp/go-ontology.html`
Get title for go-annotation:
[...]

PR #1892, opened 2022-08-16
Add GH email summary todays_date param for #1891

dustine32:
@kltm Adds an optional param `--todays_date` to set the starting date. Ex:
```
python3 github_issue_summary.py geneontology/go-ontology 1 --todays_date 2022-08-14
```
Will retrieve tickets updated or created between 2022-08-13 to 2022-08-14 (time is midnight).


kltm (approved):
Optional addition to offset the start date; looks good.

PR #1899, opened 2022-09-06
Update URLs for Pfam to point to InterPro

kimrutherford:
The Pfam website is going to be decommissioned. 

See: https://xfam.wordpress.com/2022/08/04/pfam-website-decommission/


kltm:
@kimrutherford Thank you for the heads up!

PR #1902, opened 2022-09-09
Update users.yaml

dustine32:
To complement https://github.com/geneontology/pathways2GO/issues/126.

We need some value for the contributor field for Reactome models. Without curator data available in the BioPAX, we can just use this placeholder `GOC:reactome_curators` following the SGD solution (`GOC:sgd_curators`) used for the YeastPathways contributor field.

@deustp01 @cmungall @kltm This sound good?

### DIFF
--- a/metadata/users.yaml
+++ b/metadata/users.yaml
@@ -1239,6 +1239,11 @@
   uri: 'GOC:sgd_curators'
   xref: 'GOC:sgd_curators'
 -
+  nickname: 'Curators at Reactome'
+  organization: Reactome
+  uri: 'GOC:reactome_curators'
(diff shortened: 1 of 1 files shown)


kltm:
Pausing until clarification from @vanaukenk and/or peter on next Wed meeting.

ukemi:
I'm going to put this in the list of tasks for the Reactome 82 release so it doesn't fall off the radar. We can look at it on today's call.

dustine32:
Linking the confirmation that this is good with Reactome https://github.com/geneontology/pathways2GO/issues/126#issuecomment-1246995891

Issue #1434, opened 2020-03-24
Xenbase annotations through pipeline with Noctua annotation and display / deposit


suzialeksander:
Making this ticket in Helpdesk until the block can be identified and this ticket moved to a more appropriate tracker, if needed.

XenBase has expressed interest in making use of their Noctua annotations (currently about 340 annotations). Their Noctua annotations are already available in a GPAD: http://release.geneontology.org/2020-02-26/products/annotations/noctua_xenbase.gpad.gz

Tagging @malcolmfisher103 as the contact, let me know if there's someone else at XB that should be involved.  @malcolmfisher103 can you provide any more details?

Tagging @vanaukenk and @kltm to help sort this out.

malcolmfisher103:
Thanks for that catch, we are only building our files weekly at the moment but using the daily build is definitely what we want to be doing.

We do currently have a Xenbase GAF file on our FTP but we could easily put an empty dummy one up somewhere for testing.

kltm:
Ah, you do have a GAF; could you drop the URL here?
What are the contents? Is that a derivation of the Noctua product, or something else?

malcolmfisher103:
Our GAF (http://ftp.xenbase.org/pub/DataExchange/GO/xenbase.gaf.gz) is a filtered set of annotations taken from EBI's proteomes annotation set and a GOLR query and given Xenbase format IDs for their DB Object IDs, with the original uniprot IDs ascribed as column 17 Gene Product Form IDs. It is actually an aggregation of both of these sources for both the tropicalis and laevis species.

As for the InterPro2GO annotations, all of them are coming from the EBI file as far as I can see unless there are InterPro2GO derived annotations in GOLR that don't have 'InterPro' as their 'assigned by' source or a 'GO_REF:0000002' in the reference column?

kltm:
@malcolmfisher103 Apologies for all the questions, I just want to make super sure I understand all that's going on before we look at the possibilities. What I'm trying to do is get a picture of each individual data thread so that we don't end up in a situation where one or both of us would be eating our own tail in some way.

When you're referring to GOlr, do you mean that you're pulling data from essentially AmiGO's backend, one of the GO's public endpoints?
From your description, it sounds like you GAF is from two external data locations. Could you provide more details about the API locations or download sites for your upstreams?

malcolmfisher103:
[...]

PR #1903, opened 2022-09-10
add xenbase to metadata

kltm:
Needed for #1434


metadata/datasets/paint.yaml
```diff
@@ -123,7 +123,7 @@ datasets:
    dataset: paint_other
    submitter: paint
    compression: gzip
-   source: ftp://ftp.pantherdb.org/downloads/paint/presubmission/gene_association.paint_other.gaf.gz
+   source: ftp://ftp.pantherdb.org/downloads/paint/test/15.0/2022-07-29/presubmission/gene_association.paint_other.gaf.gz
```
kltm, line 126: @dustine32 I was wondering if you remember the source/origin of this change?

PR #1915, opened 2022-10-20
Update goref-0000115.md

hattrill:
add FlyBase xref


pgaudet:
@kltm why is this failing ?

kltm:
@pgaudet That's a "symbol", not a CURIE identifier. Everything must bring along its full context.

hattrill (commented):
Have add FB: prefix
What should I do next?

kltm:
It's passing now! @pgaudet knows more context here, so it might be better for here to merge.

Issue #1924, opened 2022-11-09
Reactome id in groups.yaml should change to https


dustine32:
@ukemi [pointed out](https://github.com/geneontology/noctua/issues/798#issuecomment-1307729958) that Noctua's search function was not retrieving any models when searching by Reactome group. This is due to a "http" vs "https" discrepancy in the groups.yaml vs the Reactome models' `providedBy':
https://github.com/geneontology/go-site/blob/8cc9d3585a7b5c15bbbd9d69915a639c916dcc2e/metadata/groups.yaml#L201

@deustp01 @ukemi @kltm and I decided that "https://reactome.org" is the correct `id` value for Reactome.

I can make a quick PR.

kltm:
@dustine32 Metadata fixes are:

- [ ] https://github.com/geneontology/go-site/pull/1925

We also have data damage to deal with:
```
sjcarbon@moiraine:~/local/src/git/noctua-models/models[master]$:) grep 'http\:\/\/reactome.org' *
6258b8f300000907.ttl:	<http://purl.org/pav/providedBy> "http://reactome.org"^^<http://www.w3.org/2001/XMLSchema#string> ;
6258b8f300000907.ttl:<http://model.geneontology.org/6258b8f300000907/6258b8f300000998> <http://purl.org/pav/providedBy> "http://reactome.org"^^<http://www.w3.org/2001/XMLSchema#string> ;
6258b8f300000907.ttl:<http://model.geneontology.org/6258b8f300000907/6258b8f300001007> <http://purl.org/pav/providedBy> "http://reactome.org"^^<http://www.w3.org/2001/XMLSchema#string> ;
```
That seems pretty manually fixable, but we want to time these together probably. That would mean during an outage; this week or next time around. Is there anywhere else that this might have leaked to?

dustine32:
@kltm Thanks for finding these examples. Hm, I can't think of anywhere else this `providedBy` value could have leaked. Maybe some downstream consumer of the GO-CAM model data like if someone downloaded the whole model archive via the GO-CAM browser site?

kltm:
As the vast majority of the URIs are already "correct" with "https", I think it might be good to just go ahead and start committing these--moving ahead will make our current state less wrong than where we are now.

kltm:
For noctua, this will require a barista restart.

PR #1925, opened 2022-11-09
Update Reactome id in groups.yaml for #1924

dustine32:
For #1924


kltm (commented):
See https://github.com/geneontology/go-site/issues/1924#issuecomment-1309376319 for coordination discussion.

PR #1937, opened 2022-12-07
Update groups.yaml

suzialeksander:
adding ibb for https://github.com/geneontology/project-management/issues/43, @pgaudet please check


suzialeksander:
Also, should line 523 be PomBase not GOC? @pgaudet

kltm:
@suzialeksander Generally, we don't want a trailing '/' in the ID. `  id: https://ibeetle-base.uni-goettingen.de/` should probably be `  id: https://ibeetle-base.uni-goettingen.de`.

PR #1938, opened 2022-12-09
Add PHI-base to db-xrefs.yaml

jseager7:
This pull request adds an entry for PHI-base (the Pathogen-Host Interactions Database) to db-xrefs.yaml. I think this is needed so that PHI-base can be added to the QuickGO Assigned_by filters.

Note that the `description` states that PHI-base catalogues genes, but we actually only reuse gene and protein identifiers from other sources (Ensembl Genomes, UniProtKB, etc). The novel data we provide is on pathogen-host interactions, which is what the `entity_types` property describes: PHI:3 is an identifier for a pathogen-host interaction.

However, the next version of our database will be migrating away from pathogen-host interaction identifiers and towards gene identifiers (we might not preserve pathogen-host interaction identifiers at all), and the GAF data we've been submitting so far comes from a curation application (PHI-Canto) that is designed for this new database schema. So if the `entity_types` are intended to be cross-referenced to the GO annotations that we submit, it probably isn't ideal to use entities from what will become a legacy version of our database. Maybe none of this matters in practice, but I thought I should flag it up.

### DIFF
--- a/metadata/db-xrefs.yaml
+++ b/metadata/db-xrefs.yaml
@@ -1845,6 +1845,18 @@
   entity_types:
     - type_name: entity
       type_id: BET:0000000
+- database: PHI-base
+  name: Pathogen-Host Interactions Database
+  description: PHI-base catalogues experimentally verified pathogenicity, virulence and effector genes from fungal, Oomycete and bacterial pathogens, which infect animal, plant, fungal and insect hosts.
+  generic_urls:
+    - http://www.phi-base.org/
+  entity_types:
+    - type_name: entity
+      type_id: BET:0000000
+      id_syntax: 'PHI:[0-9]+'
+      url_syntax: http://www.phi-base.org/searchFacet.htm?queryTerm=PHI:[example_id]
+      example_id: PHI:3
+      example_url: http://www.phi-base.org/searchFacet.htm?queryTerm=PHI:3
 - database: PIR
   name: Protein Information Resource
   generic_urls:


kltm:
@pgaudet Seems technically fine, but I don't have much context here. Happy with your judgement.

ValWood:
Am merging, let me know if there is an issue...

cmungall:
If the ids are not stable then it’s best not to register this. If the goal
is to allow an assigned by then quick go should be pulling from groups.yaml

ValWood:
Do have we used the wrong yaml file?

jseager7:
I'd be happy to update `groups.yaml` with the details for PHI-base.

kltm:
@jseager7, looking at what @cmungall is saying, the path forward would be to walk back the changes in db-xrefs.yaml and add to the groups.yaml. Does this make sense for you?

kltm:
@pgaudet will ask @alexsign to clarify the source of assigned_by in quickgo.

jseager7:
@kltm I've no problem with the changes to db-xrefs.yaml being reverted. I can open a new pull request for the changes to groups.yaml if that will work with assigned_by.

kltm:
@jseager7, since the file usage is out of GO's direct sight, @pgaudet will be clarifying with @alexsign about which files are used how. You might want to hold off on any change until she can say where the change is to be applied.

pgaudet:
Hi, @alexsign confirms that this is dbxref. I'll merge this.

Issue #1943, opened 2023-01-17
Change linkouts for MOD non-human genes to go to Alliance


cmungall:
discussed on GO PIs call

This issue is resolved by making a PR on xrefs.yaml where the gene redirects go to the Alliance, not the MOD

Note: uniprot is unaffected.

pgaudet:
This is the file that should be changed: https://github.com/geneontology/go-site/blob/master/metadata/db-xrefs.yaml

kltm:
Closed in preference to newer #2499 .

Issue #1906, opened 2022-09-21
SGD links to complexes do not work 


pgaudet:
SGD annotates complexes with SGD identifiers, for example http://amigo.geneontology.org/amigo/gene_product/SGD:S000218003
However the links we provide from AmiGO are not working 
https://www.yeastgenome.org/locus/S000218003

@suzialeksander can you please check why these links are not resolving? 

Thanks, Pascale

suzialeksander:
So, the /locus/SGDID only works with locus pages.  For a complex, the URL is /complex/SGDID, so yeastgenome.org/complex/S000218003.  Can you split the SGDIDs by type?

kltm:
@suzialeksander GO policy has been that CURIEs do not contain information that needs to be parsed in the internal ID section. That would be that either SGD supplies a redirect to the right location for the general case or that the ID space is split so that the namespace is used to get things to resolve correctly. There is a similar discussion centered around https://github.com/geneontology/go-site/issues/659#issuecomment-392196555 ; TAIR and WB have gone through similar questions.

cmungall:
options:

1. create a SGD.COMPLEX namespace, analogous to what you did for pathways. bioregistry already has both:
    - https://bioregistry.io/registry/sgd.pathways
    - https://bioregistry.io/registry/sgd
    - so here would would have SGD.COMPLEX:S000218003 and this is the CURIE we would use in GO
2. have the link go to a https://www.yeastgenome.org/search?q=S000218003&is_quick=true (though the complex doesn't seem to show up in the search?)

pgaudet:
Note that the SGD complexes ID to CPX mappings should also be in the SGD GPI file
GOA cannot parse these either.

srengel:
we are requesting please to use the #2 option:

have the link go to a https://www.yeastgenome.org/search?q=S000218003&is_quick=true

pgaudet:
@kltm This is still not working as expected

kltm:
Okay, current SGD metadata is: 
https://github.com/geneontology/go-site/blob/fe5452121d10b71264c0dc58dcf07e8813843167/metadata/db-xrefs.yaml#L2372-L2415

In there, right now, there is a separate SGD_LOCUS namespace, which has the "search" link. The base SGD namespace ("SGD") uses the direct locus link.

Is the ask here that the base namespace also uses the search, or that SGD is going to start using more split namespaces. If the former, it would be as simple as adding the search link to that base namespace and doing an update.

[...]

PR #1947, opened 2023-01-25
Update db-xrefs.yaml

suzialeksander:
- do the generic_urls (http://flybase.org -> https://www.alliancegenome.org/members/flybase ?) also need to be changed? 796 1559 2320 2331 2331 2755 2862 2904

- 1566 1572 are the MGI:MGI and may not be correct

**Identifier type not in Alliance:**

- SGD no SGD_REF 
- WB no RNAi or protein page. no WB_REF
- no WBbt, WBls, WBPhenotypes
- XB morpholino oligo, transgene, line

### DIFF
--- a/metadata/db-xrefs.yaml
+++ b/metadata/db-xrefs.yaml
@@ -779,16 +779,16 @@
   synonyms:
     - FlyBase
     - FLYBASE
-  rdf_uri_prefix: http://identifiers.org/flybase/
+  rdf_uri_prefix: https://registry.identifiers.org/registry/fb
   generic_urls:
     - http://flybase.org/
   entity_types:
     - type_name: gene
       type_id: SO:0000704
       id_syntax: FBgn[0-9]{7}
-      url_syntax: http://flybase.org/reports/[example_id].html
+      url_syntax: https://www.alliancegenome.org/gene/FB:[example_id]
       example_id: FB:FBgn0000024
-      example_url: http://flybase.org/reports/FBgn0000024.html
+      example_url: https://www.alliancegenome.org/gene/FB:FBgn0000024
 - database: FBbt
   name: Drosophila gross anatomy
   rdf_uri_prefix: http://purl.obolibrary.org/obo/FBbt_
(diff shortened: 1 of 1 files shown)


suzialeksander:
For https://github.com/geneontology/go-site/issues/1943

kltm:
@balhoff I'm noticing here that some `rdf_uri_prefix`s trailing slashes and some don't; do you know what the use is here and whether that matters?

kltm:
@suzialeksander Okay, good news / bad news.
The good news is that I believe you did that URL changes correctly. :tada: 
The bad news is that they cannot work with our current system. This is part of a consistent misunderstanding of how we do CURIEs. Essentially, for this to work, we'd need a single route at the Alliance that would get CURIEed identifier to the right location
See: https://github.com/geneontology/go-site/issues/1906#issuecomment-1255426621

balhoff:
> @balhoff I'm noticing here that some `rdf_uri_prefix`s trailing slashes and some don't; do you know what the use is here and whether that matters?

@kltm I don't think trailing slashes are always needed, but I would expect some sort of delimiter (e.g., `#` or `_`).

kltm:
> > @balhoff I'm noticing here that some `rdf_uri_prefix`s trailing slashes and some don't; do you know what the use is here and whether that matters?
> 
> @kltm I don't think trailing slashes are always needed, but I would expect some sort of delimiter (e.g., `#` or `_`).

@balhoff What should these look like then for our use case? Poking around, I'm not sure where they're even getting used (and we see a mix in db-xrefs.yaml as well, originating here https://github.com/geneontology/go-site/commit/e1372ce38588cd91b712fae69a21bb7b9beea168).

kltm:
@suzialeksander @balhoff Okay, after talking to @pgaudet , I think the PR is okay to take now--I made a bunch of changes, so it would be best to look around.

@suzialeksander While the spirit of what you had is there, I did have to back out the changes to WB. There needs to be a single place for WB to go at the Alliance, just like there is with the others and like what WB has now. I don't know if we want to go ahead with what we have here or wait until we sort out the WB-specific issues (tagging @vanaukenk ).

vanaukenk:
@kltm I don't know what would have to happen at the Alliance for things to work the way they do at WB.  Is this a question for the Alliance technical team(s)?
Also, looking at the diff for this PR, is the situation with MGI the same?

kltm:
@vanaukenk @ukemi Ugh, yes--that's the same issue. I'm going to revert MGI as well.

I don't know the way forward here as I have no "specialist" PM to have contact with anymore. This is something we can look into next week?

That said, if we can't do MGI and WB for the moment, it might be more confusing than not to let this PR in.

ukemi:
Apologies for not being on board, but if I go to the file and use the URLs that were reverted they seem to resolve correctly.
For example:
https://www.alliancegenome.org/gene/MGI:1345277
https://www.alliancegenome.org/allele/MGI:3590672

However, a general MGI identifier, for example a genotype link,  doesn't work:
https://www.alliancegenome.org/allele/MGI:2167526

suzialeksander:
This does seem to be a mess, and I won't be remotely offended if this PR gets rejected and start fresh with a new one.  @kltm I'll leave that to you, let me know if you do want me to try again- I think we're waiting for a report back from the Alliance though?

kltm:
@suzialeksander I think it's probably best to bail on this PR for now and revisit after the meeting. I've converted it to "draft", so it's unlikely to get merged. Let's keep it around for a little for evidence of effort on #1943 and to give us a place to restart our work in a couple of weeks.

kltm:
Noting from @sierra-moxon that the `rdf_uri_prefix` is consumed directly from out db-xrefs.yaml by prefixmaps.

PR #1956, opened 2023-02-12
Add orcid to db-xrefs.yaml

cthoyt:
This PR adds ORCID to the GO registry. See also https://bioregistry.io/registry/orcid for more detailed information about ORCID.

This should add additional context to https://github.com/geneontology/go-ontology/pull/24694 and other efforts based on GO that want to make their contributor provenance more informative, i.e., by using standard ORCID identifiers instead of locally-defined mnemonics

### DIFF
--- a/metadata/db-xrefs.yaml
+++ b/metadata/db-xrefs.yaml
@@ -1783,6 +1783,18 @@
       url_syntax: http://www.pantherdb.org/panther/lookupId.jsp?id=[example_id]
       example_id: PANTHER:PTHR11455
       example_url: http://www.pantherdb.org/panther/lookupId.jsp?id=PTHR10000
+- database: orcid
+  name: Open Researcher and Contributor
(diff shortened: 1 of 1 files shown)


kltm:
I'm not sure about what the fix is in https://github.com/geneontology/go-ontology/pull/24694 (tagging @cmungall ), but this looks fine on this end and makes sense to have in the db-xrefs.yaml metadata file as linking (not id) information.

kltm:
Approved to run (looks like we're protected now).

cthoyt:
Thanks @kltm! Do you know if it's possible to easily rebuild http://snapshot.geneontology.org/metadata/db-xrefs.legacy to reflect this change?

kltm:
@cthoyt That rebuilds when the `snapshot` cycle completes. It typically starts every two days, but we've paused it temporarily while a release moves through the pipeline. It will likely be updated sometime next week.

PR #1967, opened 2023-02-23
Update groups.yaml

pgaudet:
Fixed agbase link


kltm:
Reverting (see skype): id, not link, so need to check first.

PR #1990, opened 2023-03-23
Update downloads-page-template.html

suzialeksander:
Changes: Fixed date in footer. Removed broken links in header- made to closely resemble header on rest of site for https://github.com/geneontology/geneontology.github.io/issues/233

lines 47-49 redundant with image link if that works

also removed               <a class="dropdown-item" href="http://amigo.geneontology.org" id="nav-crit" aria-haspopup="true" aria-expanded="false">AmiGO</a>		    
	      <a class="dropdown-item" href="http://geneontology.org/page/go-enrichment-analysis" id="nav-crit" aria-haspopup="true" aria-expanded="false">GO Enrichment Analysis</a>
 but i'd quite like to keep those esp link to AmiGO

### DIFF
--- a/scripts/downloads-page-template.html
+++ b/scripts/downloads-page-template.html
@@ -37,9 +37,8 @@
     	<span class="navbar-toggler-icon"></span>
       </button>
 
-      <a class="navbar-brand" href="index.html#home">
-    	<img src="http://geneontology.org/assets/go-logo-icon.mini.png" height="30" class="d-inline-block align-top" alt="">
-    	Gene Ontology Consortium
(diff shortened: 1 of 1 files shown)


suzialeksander:
@kltm if you want to look this over, I tried not to touch any parts of the page that are functional for downloads, just the header/footer.  It's not perfect yet my goal is just to not have made anything worse

kltm:
@suzialeksander Eyeballing it, it looks good to me. If you make sure it looks okay when you're done, I think you can merge it in whenever you want.

Issue #1992, opened 2023-03-30
Add species_code to YAMLs


suzialeksander:
The following yamls are missing at least one species_code, and probably should have them: 

cgd.yaml
ecocyc.yaml

genedb
noctua 
paint

Sounds like  /scripts/downloads-page-gen.py is looking for this line and hoping it's a step to fix https://github.com/geneontology/go-site/issues/1259

kltm:
I believe this is done?

PR #1994, opened 2023-04-04
Suzialeksander patch 31

suzialeksander:
Adding species codes for https://github.com/geneontology/go-site/issues/1992


kltm:
I'm genuinely not sure that noctua.yaml and paint.yaml has much effect, but it should be fine to have the metadata there.

kltm (approved):
Looks good to me.

Issue #1259, opened 2020-01-03
Update Downloads table 


pgaudet:
Need to update certain files to show the correct species: 

http://current.geneontology.org/products/pages/downloads.html

Multi-species
Candida Genome Database (cgd) cgd.gaf 
Multi-species
Encyclopedia of E. coli metabolism (ecocyc) 	ecocyc.gaf 
Multi-species
GeneDB (genedb) 	genedb_pfalciparum.gaf 
Multi-species
GeneDB (genedb) 	genedb_lmajor.gaf 
Multi-species
GeneDB (genedb) 	genedb_tbrucei.gaf 
Multi-species
Plant-Associated Microbe Gene Ontology Interest Group (pamgo) 		pamgo_ddadantii.gaf
Multi-species
Plant-Associated Microbe Gene Ontology Interest Group (pamgo) 	pamgo_atumefaciens.gaf
Multi-species
Plant-Associated Microbe Gene Ontology Interest Group (pamgo) 	pamgo_oomycetes.gaf 

1. Those corresponds to single species, not multi species
2. Now that we have changed the source of these files (GOA rather than PAMGO/GeneDB), we need to make sure that these files are still correctly generated @kltm 

Thanks, Pascale

kltm:
@dougli1sqrd @pgaudet I think these are all generated from the metadata, right?

pgaudet:
@kltm @dougli1sqrd @lpalbou Which file do I need to edit to fix this ? ie where is the "Species" label coming from ? 

Thanks, Pascale

suzialeksander:
Also double check these are fixed when this ticket is addressed: 

> Multi-species
Encyclopedia of E. coli metabolism ([ecocyc](http://ecoliwiki.org/))

- [ ] Needs to be update to E. coli encyclopedia 
- [ ] Remove multispecies

suzialeksander:
_This is very much a multi-species file, all yeast but not all Candida:_
~~Multi-species~~
~~Candida Genome Database (cgd) cgd.gaf~~

_Waiting for next release to see if these clear:_ 
Multi-species
Encyclopedia of E. coli metabolism (ecocyc) 	ecocyc.gaf 
Multi-species
GeneDB (genedb) 	genedb_pfalciparum.gaf 
Multi-species
GeneDB (genedb) 	genedb_lmajor.gaf 
Multi-species
GeneDB (genedb) 	genedb_tbrucei.gaf 

_These aren't in our downloads anymore:_
[...]

PR #1995, opened 2023-04-04
Update downloads-page-gen.py

suzialeksander:
Adding species mappings to address https://github.com/geneontology/go-site/issues/1259

Not sure if the recent (Feb-March) changes with GeneDB should/can be issued here, but the goal of this PR is just to fix the Multi-species label


kltm:
I forget what this all is, but the updates look good.

PR #1997, opened 2023-04-04
Update downloads-page-gen.py

suzialeksander:
removing Calb as that is no longer referenced in any yaml

### DIFF
--- a/scripts/downloads-page-gen.py
+++ b/scripts/downloads-page-gen.py
@@ -27,7 +27,6 @@ smap = {
     "Aspergillus": "Aspergillus nidulans",
     "Atal": "Arabidopsis thaliana",
     "Btau": "Bos taurus",
-    "Calb": "Candida albicans"
     "Cele": "Caenorhabditis elegans",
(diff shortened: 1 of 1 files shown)


pgaudet:
Hi @suzialeksander 

https://github.com/geneontology/go-site/blob/master/metadata/datasets/cgd.yaml
mentions TaxID 5476

[Candida albicans](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Tree&id=5476&lvl=3&keep=1&srchmode=1&unlock)

suzialeksander:
@pgaudet Yes, to clarify: I added `Calb` to the cgd.yaml with a mapping on this script as (in theory) that would then be pulled to replace the "Multi-species" label on the Downloads page http://current.geneontology.org/products/pages/downloads.html.  But I later realised it would be incorrect to call CGD's GAF "C albicans" as there are other yeasts in there too, I've removed this mapping and it will remain multi-species

pgaudet:
Ah OK thanks for the clarification!

PR #1998, opened 2023-04-05
Update ecocyc.yaml

suzialeksander:
@kltm just checking- this `compression:` field refers to the file in the `url:` field, not the `source:`, correct?  Source file is not .gz

Adding this as the field on products/pages/downloads.html is currently blank


kltm:
@suzialeksander It is very much referring to the source, not the "url"--you'll need to remove that addition (or just close this PR) as having that would prevent the pipeline from working properly.

kltm:
All files we create are uniform, so there is no point in having metadata related to our final products.

PR #2007, opened 2023-05-18
Issue 1519 exclude annots with source goc

mugitty:
Updated rule description


kltm:
@mugitty Text-only changes. Feel free to merge.

PR #2009, opened 2023-05-22
Update db-xrefs.yaml

suzialeksander:
Adding new YeastPathway IDs for https://github.com/geneontology/pathways2GO/issues/254

@kltm I'm not sure if `type_name: entity` and `type_id: BET:0000000` are the correct choices but I didn't see anything better?


kltm:
@suzialeksander If it's consisten with usage elsewhere, it's probably fine. We technically do not use it locally, so I'm not too concerned personally, but it still may be used at GOA.

pgaudet:
@alexsign doesn't foresee any problem for GOA, please go ahead and merge.

PR #2017, opened 2023-05-26
Revise SPARQL queries that mark Noctua models and filter non-production.

### LINKED ISSUE #2013: production triplestore includes non-production models
I think the filtering query is getting confused due to `prov:wasDerivedFrom` now being also used to connect copied models to their sources.

balhoff:
Fixes #2013.

With this change, asserted Noctua models continue to have `graphType` `noctuaCam`. Now the graphs containing inferences have `graphType` `noctuaInferences` instead of `noctuaCam`.


kltm (approved):
From my reading, these changes seem good.
Feel free to merge on your schedule and you can then check the product on the next `snapshot` output.

PR #2024, opened 2023-06-20
Update fb.yaml

hattrill:
Update of FB file with gpi

### DIFF
--- a/metadata/datasets/fb.yaml
+++ b/metadata/datasets/fb.yaml
@@ -22,4 +22,18 @@ datasets:
    species_code: Dmel
    taxa:
     - NCBITaxon:7227
- 
+ -
+   id: fb.gpi
+   label: "fb gpi file"
+   description: "gpi file for Drosophila melanogaster from FlyBase"
+   url: http://current.geneontology.org/annotations/fb.gpi.gz
+   type: gpi
+   dataset: fb
+   submitter: fb
+   compression: gzip
+   source: http://ftp.flybase.net/releases/current/precomputed_files/collaborators/gp_information.fb.gz
+   entity_type:
+   status: active
+   species_code: Dmel
+   taxa:
+    - NCBITaxon:7227


hattrill:
@kltm this file is also used by Protein2GO. Wouldn't be able to change for this release. Could change for next, which would be a couple of months. Can you use this one for now?

kltm:
@hattrill I think remote-controlling P2GO through our application metadata is out of scope for this file. While other groups incidentally using this is fine (and for db-xrefs.yaml encouraged), the only current purpose of this entry is to signal what file to use for the NEO (Noctua autocomplete) entity load.

kltm:
If we want to have a secondary product for external consumption about what our upstreams are, we could add that as a release data product of have a secondarily generated product.

hattrill:
Hi @kltm 

This gpi pointer was added to the yaml after the email correspondance saying that it was all fine to use the file that we use for P2GO lookup for Noctua. Not seeing what the issue is as you just need FBgn-symbol-name correspondance and can ignore the extra UniProt ID mapping column.


Thread:
Re: gpi
started on 20/05/2023
"Hi Kimberly, Pascale,

We currently don't submit a gpi file to GO and so our IDs in Noctua are
generated from the GAF. We make a gpi for P2GO - attached. I think that
this should serve just as well for Noctua. What do you think?

Thanks,

Helen"

"Hi Helen,

 

(cc-ing Seth so he is in the loop) – this is great! Please let me know where we can get the file, I will update the FB yaml file to include this information[....]

 Thanks, Pascale"

"Lovely. I am going to get this file on our public ftpsite so it will be there by mid-June. I can edit the yaml file when it's ready.

Thanks,

Helen"

kltm:
@hattrill Perfect--that's exactly the information I was looking for. I just wanted to confirm that the semantics of entries in the metadata file were understood, as they are driving our pipeline directly and that is the necessary driving purpose of the file. (I also don't have a great memory, so thank you for the reminder that I likely looked as this a bit back.)

cmungall:
Why are the genes typed as "proteins"?

pgaudet:
Do you mean it should be protein-coding gene'?

PR #2048, opened 2023-08-25
prepare metadata for goa mirror

kltm:
From geneontology/pipeline#305

### DIFF
--- a/metadata/datasets.schema.yaml
+++ b/metadata/datasets.schema.yaml
@@ -112,6 +112,10 @@ mapping:
             type: str
             required: false
             unique: false
+          "mirror_of":
+            type: str
+            required: false
+            unique: false
           "species_code":
             type: str
             required: false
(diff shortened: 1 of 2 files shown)


dustine32 (approved):
I think this is good to test. So, `ontobio` will start downloading from https://mirror.geneontology.io/ (for goa at least) since it always uses the `source` field and of course there _should_ always be some file at that `source` location.

kltm:
@dustine32 Correct: anything using the `source` field in this metadata file will start grabbing from the mirror.
The mirror is populated in a very basic way from the [goa-copy-to-mirror](https://build.geneontology.org/job/geneontology/job/pipeline/job/goa-copy-to-mirror/) branch, moving files to the `go-mirror` bucket, which then has Cloud _Front_ producing mirror.geneontology.io.
That said, I think the new location should only be an improvement for most practical purposes. I'm not sure how to test here besides letting the pipelines at the new files.

dustine32:
@kltm Oh sorry, I guess by "test" I meant let them run and in the unlikely event it doesn't work we'll change back (there's probably a better word).

PR #2050, opened 2023-08-29
add requirements.txt to scripts directory in order to install it in t…

sierra-moxon:
…he pipeline stage where teh scripts are used instead of a requirements.txt in the docker image itself.

### DIFF
--- a/scripts/requirements.txt
+++ b/scripts/requirements.txt
@@ -1,0 +1,5 @@
+PyYAML
+ontobio==2.8.4
+click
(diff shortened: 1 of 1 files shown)


kltm (approved):
New file, should not disrupt current "raw" uses; needs to be run through new dev-base image.

kltm:
I'm okay with and merge timing.

dustine32 (approved):
Plan is to merge into test branch `issue-844-ext-repair-obs-terms`

PR #2096, opened 2023-10-07
Roll back ontobio hiccup

kltm:
Tagging @dustine32


dustine32:
@kltm Yup, I'm down with this reversion. I found the reason for the test fail and the fix will be in `ontobio==2.8.16`.

dustine32 (approved):

Issue #2002, opened 2023-04-19
gorule-0000061 should be displayed in yellow


pgaudet:
http://release.geneontology.org/2023-04-01/reports/gorule-report.html

gorule-0000061 should be displayed in yellow, since it's a warning, not a filter.

pgaudet:
@mugitty 

Not clear why this is not being repaired: 
UniProtKB P67881 CYC colocalizes_with GO:0043293 PMID:12220121 IDA C Cytochrome c CYC protein taxon:9031 20110427 AgBase 

in http://snapshot.geneontology.org/reports/assigned-by-AgBase-report.html#gorule-0000061
(snapshot Generated on 2023-09-20)

Thanks, Pascale

pgaudet:
Working on snapshot

pgaudet:
to check in the next (or next-next) release

pgaudet:
Looks OK on latest (Oct 23) release
http://release.geneontology.org/2023-10-09/reports/assigned-by-gorule-report.html

PR #2097, opened 2023-10-07
Update requirements.txt

dustine32:
This ups the `ontobio` to pull in the [fix](https://github.com/biolink/ontobio/pull/649) for #2002 specifically to handle obsoleted terms in the GO rule 61 QC test.


kltm (approved):
Looks good, if we pass the tests.

Issue #2093, opened 2023-10-06
gorule-0000030 test is not failing 


pgaudet:
It seems gorule-0000030 is not implemented correctly; the test in the[ test GAF](https://github.com/geneontology/go-site/blob/master/docs/gorules_test_errors.gaf) is not failing 
Added a test (TEST 2) with GO_REF:0000033; it may be that this is the only GOREF tested (see https://github.com/geneontology/go-site/issues/1118)

* Need to know if the **obsoleted status** (which is a boolean value in the go_ref file) of the reference is checked in the [goref file](https://github.com/geneontology/go-site/tree/master/metadata/gorefs) or if it is [hard coded in the rule](https://github.com/geneontology/go-site/blob/master/metadata/rules/gorule-0000030.md) 
*

pgaudet:
Once tests fail correctly, consider implementing this rule as an error, see https://github.com/geneontology/go-site/issues/1423

pgaudet:
Ontobio code fixed, see https://github.com/biolink/ontobio/pull/653#pullrequestreview-1695979623

pgaudet:
tests not failing

pgaudet:
Looks like this is working, but that we're not using the latest versions of the GOREFs, since GOREF_0000033 is returning an error, but it has been un-obsoleted

pgaudet:
Now looks OK in the test GAF.

PR #2127, opened 2023-10-19
For #2093

mugitty:


dustine32 (changes requested):
Straightforward change of just supplying the new metadata param. Though this is dependent on whichever `ontobio` release contains https://github.com/biolink/ontobio/pull/653 so, this PR should also contain the [pipeline/requirements.txt](https://github.com/geneontology/go-site/blob/master/pipeline/requirements.txt) `ontobio` bump change once that is released.

dustine32 (approved):
Updated `reqs.txt` so great!

PR #2192, opened 2023-11-27
Create gorule-0000012.md

pgaudet:


pgaudet:
Same as [GORULE:0000016](https://github.com/geneontology/go-site/blob/master/metadata/rules/gorule-0000016.md)

PR #2218, opened 2024-01-17
Update requirements.txt for GO_REF:0000033

dustine32:
This `ontobio` bump will pull in the correct check for `GO_REF:0000033` in IBAs - https://github.com/biolink/ontobio/pull/665


kltm:
Bumpity-bump.

PR #2231, opened 2024-02-12
Change TC to TCDB.

cmungall:
See also https://github.com/geneontology/go-ontology/issues/27025

Note: this can be merged in advance of https://github.com/linkml/linkml/pull/1907 because I kept TC as a synonym


kltm:
@cmungall Is this ready; do you want a review from @balhoff ?

PR #2259, opened 2024-03-14
Update gorefs.yaml

pgaudet:
For https://github.com/geneontology/gopreprocess/issues/5


pkalita-lbl (approved):
Looks great! Thanks for keeping the YAML file in sync with the Markdown files while we're in this transition period.

Issue #2273, opened 2024-03-27
Proposed  new gorule-0000064 TreeGrafter IEAs should be filtered for GO reference species


cmungall:
Proposed text:

If an annotation is an IEA to GO_REF:0000118 and the taxon is not in the GO reference species list (i.e 143 species for which we get IBAs), then filter out.

Ideally this would be implemented upstream at EBI-GOA

We have a goref for tree grafter #2130 : [GO_REF:0000118](https://github.com/geneontology/go-site/blob/master/metadata/gorefs/goref-0000118.md)

pgaudet:
Test errors are correctly being picked up.

pgaudet:
Working as expected.

pgaudet:
Hi @mugitty 

Tests are not failing - or not being reported in error reports.

mugitty:
@pgaudet, tests are failing for gorule 64 in snapshot:
http://snapshot.geneontology.org/reports/assigned-by-TreeGrafter-report.html#gorule-0000064
...
gorule-0000064

TreeGrafter IEAs should be filtered for GO reference species

    total: 10050

messages

    ERROR - Violates GO Rule:GORULE:0000064: TreeGrafter ('GO_REF:0000118') IEAs should be filtered for GO reference species--`SGD S000005513 YOL153C involved_in GO:0051603 GO_REF:0000118 IEA PANTHER:PTN000110795 P Blocked reading frame, contains two in-frame stops in translation YOL153C protein taxon:559292 20240610 TreeGrafter UniProtKB:P0C155 `
    ERROR - Violates GO Rule:GORULE:0000064: TreeGrafter ('GO_REF:0000118') IEAs should be filtered for GO reference species--`SGD S000005513 YOL153C located_in GO:0000328 GO_REF:0000118 IEA PANTHER:PTN000110795 C Blocked reading frame, contains two in-frame stops in translation YOL153C protein taxon:559292 20240610 TreeGrafter UniProtKB:P0C155 `
...

In the errors.gaf file processing, we have the following:
! FAILS GORULE:0000064 - TEST 1 - human protein
UniProtKB	A0A140VJL9	A0A140VJL9	involved_in	GO:0006783	GO_REF:0000118	IEA	PANTHER:PTHR11458:SF0	P	GORULE_TEST:0000064-1		protein	taxon:9606	20240201	TreeGrafter	
! FAILS GORULE:0000064 - TEST 2 - Drosophila protein
[...]

PR #2282, opened 2024-03-28
Bump ontobio for adding gorule-0000064

dustine32:
For #2273. This brings in change https://github.com/biolink/ontobio/pull/672 to filter out IEAs with TreeGrafter `GO_REF:0000118` for species in [go-reference-species.yaml](https://github.com/geneontology/go-site/blob/master/metadata/go-reference-species.yaml).


kltm:
All tests passed! Thank you @mugitty and @dustine32 !

Issue #2235, opened 2024-02-14
Trigger deploy workflow of `geneontology.github.io` when `gorefs.yaml` changes


pkalita-lbl:
To be done after https://github.com/geneontology/geneontology.github.io/issues/507

- [ ] Add a new workflow to this repo which is triggered on changes to `metadata/gorefs.yaml` in the `master` branch. The workflow should use the [GitHub CLI](https://cli.github.com/manual/) to trigger the build and deploy workflow (created as part of https://github.com/geneontology/geneontology.github.io/issues/507) in the `geneontology.github.io` repo (something along the lines of `gh workflow run -R ...`). We'll need a repo administrator to add a GitHub token with permission to run that workflow as a secret in this repo. We'll also need to take care to coordinate this with the QC workflow and ensure that this workflow only runs if QC succeeded.

kltm:
@pgaudet @suzialeksander When we reach this point, it would be good to review the changes (minimal to none at your end) and think about how (if?) to link this from the main geneontology.org pages.

pgaudet:
It may be a good idea to deploy these on the GO website, unless the layout we have improves; right now it's a bit weird to send users to the GH markdown page (for eg https://github.com/geneontology/go-site/blob/master/metadata/gorefs/goref-0000001.md) - if I were a user I would not be sure whether I landed there on purpose.

kltm:
@pgaudet To clarify, have these built in to the main site is the point of those section. I'm trying to specify how you want to specifically present it.

pgaudet:
You comment " think about **how (if?)** to link this from the main geneontology.org pages." 

sounded like we were considering moving the GOREFs to the website, not that it was decided. Thanks for the clarification.

pkalita-lbl:
Yes, just reiterating that the plan is to deprecate and [remove](https://github.com/geneontology/go-site/issues/2237) the Markdown files. There will be pages instead of geneontology.org. The format of those pages isn't fully specified. As part of https://github.com/geneontology/geneontology.github.io/issues/506 I imagine I will make a prototype of those pages and get feedback before finalizing it.

pkalita-lbl:
This is now done with https://github.com/geneontology/go-site/pull/2294 and related PRs.

[...]

PR #2294, opened 2024-04-12
Trigger website rebuild when gorefs.yaml changes

pkalita-lbl:
Re: https://github.com/geneontology/go-site/issues/2235 -- I will close the issue manually when I'm confident this is working as expected.

### Summary

These changes add a new workflow file (`.github/workflows/trigger-website-rebuild.yaml`) which runs when the `metadata/gorefs.yaml` file is changed in the `master` branch. 

It first validates the `gorefs.yaml` file according to its associated schema. Theoretically any validation issues would have been caught by the `qc.yaml` workflow during a PR. So this is just sort of a fail-safe against someone ignoring that check or committing directly to `master`. 

If the validation passes we generate an access token using the GO Workflow Bot app installed in this repo. Then we use that token and the GitHub CLI[^1] to trigger the `deploy.yaml` workflow[^2] in the `geneontology/geneontology.github.io` repo.

[^1]: https://docs.github.com/en/actions/using-workflows/using-github-cli-in-workflows
[^2]: https://cli.github.com/manual/gh_workflow_run

### DIFF
--- a/.github/workflows/trigger-website-rebuild.yaml
+++ b/.github/workflows/trigger-website-rebuild.yaml
@@ -22,6 +22,7 @@ jobs:
       # In theory, a failed validation would have been caught by the qc workflow before this point.
       # But we validate again here just to double-check that we're not triggering a rebuild based
       # on invalid data. This is also simpler than tying this workflow to the qc workflow.
+      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


kltm (approved):
Added workflow as discussed. Seems good!

PR #2298, opened 2024-04-17
Rollback some upstream files to last good

dustine32:
For https://github.com/geneontology/pipeline/issues/373.

This changes `source:` in these four datasets YAMLs to point to their `*-src.gaf.gz` files in https://release.geneontology.org/2024-03-28/products/upstream_and_raw_data/. This PR is necessary to avoid changes in their current upstream files. I'll provide some simple file length comparisons (`wc -l`) of rollback 2024-03-28 versions vs the current upstreams:

Dictybase - 81166 (rollback) vs 50732 (current)
MGI - 272742 (rollback) vs 310193 (current)
PomBase - 46162 (rollback) vs 45589 (current)
TAIR - 235092 (rollback) vs 189228 (current)

Tagging @kltm @pgaudet

### DIFF
--- a/metadata/datasets/dictybase.yaml
+++ b/metadata/datasets/dictybase.yaml
@@ -52,7 +52,7 @@ datasets:
    dataset: dictybase
    submitter: dictyBase
    compression: gzip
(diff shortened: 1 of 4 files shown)


kltm (approved):
Changes to four upstreams as requested.

PR #2302, opened 2024-04-25
Update wb.yaml

vanaukenk:
Changing file source back to WB ftp which seems to be back up.


kltm:
Cheers!

kltm:
https://github.com/geneontology/pipeline/issues/373

PR #2306, opened 2024-04-29
Final removal of GO_REF Markdown files

### LINKED ISSUE #2237: Clean up remants of GO_REF Markdown files
Once https://github.com/geneontology/go-site/issues/2236 is complete I don't believe there will be any reason to hang on to the old Markdown files anymore. We should remove them from this repo. We can probably also jettison the scripts that were previously used to condense the Markdown files into a YAML file.

pkalita-lbl:
Fixes #2237 

### Summary

This is it. The final goodbye to GO_REF Markdown files. Thus ends the transition period where GO_REFs were defined in both the old Markdown files and the new YAML file. It's only the YAML file from here on out.

### Details

* One last refresh of `metadata/gorefs.yaml` to get any updates that were only in the Markdown files.
* Removed the files in `scripts/goref_parser`, which had been used to do the Markdown-to-YAML conversion
* Removed `scripts/aggregate-references.py`, which had been used to generate `GO.references` and `go-refs.json` from the Markdown files
* Removed `scripts/bootstrap-gorefs-md.pl`, which was used to create the Markdown files from `GO.references`
* Removed `scripts/generate-gorefs-summary.pl`, which was used to generate `metadata/gorefs/README.md` from the individual GO_REF Markdown files.
* Updated miscellaneous links to point to either the new YAML file or the https://geneontology.org/gorefs index page, depending on what felt appropriate
* **Removed the `metadata/gorefs` directory**

### Comments

* If we decide that we really want to continue supporting the `GO.references` [file](http://current.geneontology.org/metadata/GO.references), let me know and I can un-delete and rewrite `scripts/aggregate-references.py` to work off of the new YAML file instead.
* If we want to continue supporting the `go-refs.json` [file](http://current.geneontology.org/metadata/go-refs.json), we'll need to have the pipeline convert the YAML file into JSON, at minimum. However, **the YAML file does not use exactly the same field names as the existing JSON file** (e.g. `description` vs `abstract`). If we want to exactly match the existing JSON file format, we'll need to do more than a simple YAML-to-JSON conversion in the pipeline.
* I have a vague sense of unease about the change here to `pipeline/go-association-transit/goat/parse_into_gpad.py`. It seems to imply that Python code in the `ontobio` package has an implicit understanding of the old (and now removed) `metadata/gorefs` directory. I took a brief look through that code, but it wasn't immediately clear to me how much might break in `ontobio` with the removal of the `metadata/gorefs` directory. I'm hoping that @kltm can advise if this is something that needs to be resolve before we merge these changes or not.


kltm:
@mugitty @dustine32 Would one of you be able to take a look at @pkalita-lbl 's reference above to parse_into_gpad.py and see if there is any need to worry? I'm not sure goat was every used...

pgaudet:
Nice work @pkalita-lbl !

kltm:
Okay, I think we're clear to bring this in; however, as it will break the pipeline, I want to wait until the current snapshot run has ended.

pkalita-lbl:
> I want to wait until the current snapshot run has ended

Sounds good. Do you want to let me know when that's happened? Also do you want me to take care of removing [the call](https://github.com/geneontology/pipeline/blob/30108ded53e95cadc05f9615dcd522f7ecb62714/Jenkinsfile#L831) to `aggregate-references.py` from the pipeline? Or is there more to it than I realize?

kltm:
> Sounds good. Do you want to let me know when that's happened? Also do you want me to take care of removing [the call](https://github.com/geneontology/pipeline/blob/30108ded53e95cadc05f9615dcd522f7ecb62714/Jenkinsfile#L831) to `aggregate-references.py` from the pipeline? Or is there more to it than I realize?

No worries, I can take care of that and merge the PR when we're ready.

pkalita-lbl:
Works for me! Thanks

kltm:
@pgaudet @pkalita-lbl , The main pipeline (at this point `snapshot-post-fail`) should be protected now (although some branches may still need line removed @dustine32 @sierra-moxon).
I'm going to go ahead and merge.

kltm (approved):
A lot of changes (removals) but changes make sense to me.

PR #2317, opened 2024-05-02
Issue 2281 only implemented rules in reports

mugitty:
Output information in reports only for rules that have been implemented

### DIFF
--- a/scripts/reports-page-gen.py
+++ b/scripts/reports-page-gen.py
@@ -34,7 +34,6 @@ def main(report, template, date, suppress_rule_tag):
     for rule_path in glob.glob(os.path.join(rules_directory, "gorule*.md")):
         with open(rule_path) as rule_file:
             rule = yamldown.load(rule_file)[0]
-
             rule_id = rule["id"].lower().replace(":", "-")
             rules_descriptions[rule_id] = {
                 "title": rule["title"],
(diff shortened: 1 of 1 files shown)


kltm:
@mugitty I just want to confirm a bit of overall logic: rules that have been implemented elsewhere (i.e. not ontobio), are still "implemented", but would not appear in the reports from ontobio, but may appear in other reports and may be merged into the main reports some other way. Naturally, no un-implemented things should appear in any reports, but implemented things not in ontobio may.

mugitty:
@kltm, do we want ontobio to check if the status of the rule is approved or implemented?

mugitty:
The MODS use the reports to check for errors. @pgaudet did not want errors to be displayed before the functionality for the checks was implemented

kltm:
@mugitty Again, that's all fine. I just want to be clear on GORULEs not being strictly limited to ontobio. This all seems fine within the expected scope.

kltm (approved):
Simple changes to filter ontobio-local rules.

pgaudet:
Hi @kltm 

Thanks for the feedback. In fact I had asked Eric to have the information of 'where are rules implemented' avialable in the GORULEs info - . I think that corresponds to "implementations" in the [schema](https://github.com/geneontology/go-site/blob/master/metadata/rules.schema.yaml) . Values here should be an enum. 

This information would be very useful.

PR #2326, opened 2024-05-08
Update group-contacts.csv

nsuvarnaiari:
Removed Becky Tauber and James Munro from ECO's contact list


pgaudet:
I think theere was a conflict - redid.

PR #2338, opened 2024-06-27
move database name from taxon->NCBITaxon

sierra-moxon:
solves part of the issue here: https://github.com/geneontology/go-fastapi/issues/95 (contracting URI -> CURIE for NCBITaxon identifiers results in `taxon:` CURIE prefix instead of `NCBITaxon:` CURIE prefix).

The other way to handle this is to special-case this prefix in the prefixmaps repo.


kltm:
Heads up to @tmushayahama @dustine32 @pkalita-lbl , I'm _assuming_ this won't break anything, and if it does we should fix it as this is more correct, but I want to make everybody aware for breakages around this when it goes live with a release or API rollout.

kltm (approved):
Seems reasonable, with the usually worrying caveats.

PR #2340, opened 2024-07-02
enable testing on snapshot of MGI upstream GAF produced at GO.  

sierra-moxon:
PR  changes MGI metadata to retrieve upstream from MGI ingest pipeline and adds the GPAD/GPI 2.0 flag to validate.py produce command so that we start outputting GPAD/GPI 2.0 in snapshot. 

This should not be merged until we have a clean snapshot build.

### DIFF
--- a/metadata/datasets/mgi.yaml
+++ b/metadata/datasets/mgi.yaml
@@ -26,7 +26,7 @@ datasets:
    dataset: mgi
    submitter: mgi
    compression: gzip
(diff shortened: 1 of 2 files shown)


kltm (approved):
Exciting!
I have one last bug I'm running down in the second stage. We can merge this as soon as the next `snapshot` clears.

kltm:
`snapshot` cleared

kltm:
@sierra-moxon was this tied to a particular issue for the project?

PR #2343, opened 2024-07-11
correct mistaken --gpad removal when adding gpad/gpi version output f…

sierra-moxon:
…lag set to 2.0


kltm:
Following on https://github.com/geneontology/go-site/pull/2340

kltm (approved):
Makes sense to me!

PR #2356, opened 2024-08-14
bump ontobio version to get in isoform fixes

sierra-moxon:


kltm (commented):
This looks good, but that error in the tests would seem to be a problem

kltm:
Any ideas here @dustine32 ? I think this is the first time I've seen this...

dustine32:
I only see as high as `2.9.3` in [PyPI](https://pypi.org/project/ontobio/#history), not `2.9.5`. Maybe it's just the usual few minutes of lag between when the version is pushed and PyPI displays it?

dustine32:
@sierra-moxon I see the two missing versions (2.9.4, 2.9.5) in the `ontobio` [releases page](https://github.com/biolink/ontobio/releases). Maybe these releases somehow didn't actually push to PyPI?

sierra-moxon:
yeah as per slack, I'm just getting a GH action in place in ontobio repo to push those to pypi on "release".

sierra-moxon:
pypi GH action worked in ontobio, bumping the available version in pypi to v2.9.6.

sierra-moxon:
@kltm - alrighty - after reverting the autopush, manual push to pypi worked well and tests here are now passing.  ready when you are to try this in snapshot.

kltm (approved):
Now passing

PR #2362, opened 2024-09-06
bump ontobio to 2.9.8

sierra-moxon:
would like to try this ontobio release on snapshot again, to verify a fix for https://github.com/biolink/ontobio/issues/689


kltm:
While we won't get any final product out until we can clear https://github.com/geneontology/sparql-for-pathway-go-cams/issues/3, this looks good to me for trialling.

kltm (commented):
simple bump

PR #2365, opened 2024-09-25
Output GPAD 2.0 from temp post filter

dustine32:
For https://github.com/geneontology/pipeline/issues/386

Hopefully, this plugs the GPAD 1.2 leak we're seeing in the pipeline testing for 2.0.

Tagging @kltm @sierra-moxon


kltm (approved):
Adding explicit content.

PR #2366, opened 2024-09-26
Add db-xref config for PomBase transcripts

kimrutherford:
Refs geneontology/helpdesk#503


kltm:
@kimrutherford Seems consistent to me.

PR #2370, opened 2024-10-07
Add rdf_uri_prefix for PMID and GO_REF

pkalita-lbl:
Related to https://github.com/geneontology/gocam-py/issues/13

These changes add a `rdf_uri_prefix` field for PubMed and GO_REF databases. This is with an eye toward having them included in the `go` context of [`prefixmaps`](https://linkml.io/prefixmaps/index.html).

### DIFF
--- a/metadata/db-xrefs.yaml
+++ b/metadata/db-xrefs.yaml
@@ -971,6 +971,7 @@
       type_id: BET:0000000
 - database: GO_REF
   name: Gene Ontology Database references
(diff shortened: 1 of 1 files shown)


kltm (approved):
This looks right to me.
I think this is fine to merge on your timing, but I'd also just like to make sure this is on @balhoff 's radar.

balhoff:
Good as far as I know!

Issue #2374, opened 2024-10-15
COG url_syntax


carrere:
Dear GeneOntology team,

Concerning your Xref reference page [https://amigo.geneontology.org/xrefs]

The correct URL Syntax for COG Cluster seems to be (with prefix COG: in various annotation pipelines)

```
url_syntax: https://www.ncbi.nlm.nih.gov/research/cog/cog/[example_id]
```

Sincerely yours,

suzialeksander:
@carrere thank you for reporting this!

tagging @kltm

pgaudet:
Thanks @carrere 

I fixed this in our dbxref file. 

I notice we have two more COG databases: COG_Function and COG_Pathway

The links are respectively: http://www.ncbi.nlm.nih.gov/COG/grace/shokog.cgi?fun=H
and http://www.ncbi.nlm.nih.gov/COG/new/release/coglist.cgi?pathw=14
but are not working. Do you have newer URLs for these? 

Thanks, 

Pascale

carrere:
Hi @pgaudet ,

I'm not working at NCBI but I think correct URLs are:

- COG_Function: https://www.ncbi.nlm.nih.gov/research/cog/cogcategory/[example_id]  (eg: [For J class function](https://www.ncbi.nlm.nih.gov/research/cog/cogcategory/J)
- COG_Pathway: It is a name of pathway and not a 
-  number like in your example; https://www.ncbi.nlm.nih.gov/research/cog/pathway/**Isoprenoid%20biosynthesis**/
Bests,
Sebastien

PR #2375, opened 2024-10-17
Update db-xrefs.yaml for COG_Cluster

kltm:
Update URL syntax for COG_Cluster; for #2374


kltm:
@suzialeksander Noting that all other COG-related URLs likely also incorrect.

PR #2389, opened 2024-11-20
Update users.yaml

pgaudet:
Off boarding Federica Quaglia 
Change edit rights in Noctua 
@kltm is this enough to remove her access to Noctua? Or do you need to change something else?

### DIFF
--- a/metadata/users.yaml
+++ b/metadata/users.yaml
@@ -4550,7 +4550,7 @@
   authorizations:
     noctua:
       go:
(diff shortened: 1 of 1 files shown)


kltm:
This would be analogous to what we do when we add a new user: you want to probably tag @vanaukenk on the PR and have her do a merge and user permissions refresh when the timing is right.

kltm (approved):
This looks right to me.

PR #2395, opened 2024-12-12
Update dictybase.yaml

pgaudet:


pgaudet:
@kltm Can you see why this is failing ? It looks like examples are being validated by the checks?  is this new?

kltm:
@pgaudet This is not new. I believe you added new examples for 0000065 yesterday that did not pass (https://github.com/geneontology/go-site/commits/master/metadata/rules/gorule-0000065.md), but was merged.

PR #2428, opened 2025-01-31
add GPIs for 50 vir/bac species

kltm:
For geneontology/neo#116; this adds the (hopefully inactive) metadata necessary for the new request GPIs to filter into the NEO build automagically.

The question here is whether there would be unintended consequences.


kltm:
In discussion on dev slack, we were unable to reason how this might go wrong. Will proceed tentatively.

Issue #2418, opened 2025-01-24
Small formating issue


pgaudet:
Hi @mugitty 

In individual report pages, for example 
https://release.geneontology.org/2024-11-03/reports/assigned-by-dictybase-report.html

the 'quotes' resolve as ASCI codes

<img width="1089" alt="Image" src="https://github.com/user-attachments/assets/bf776292-12e6-48c2-8321-a53f7472e5cc" />

If this is an easy fix, can it be made prettier? 

Thanks, Pascale

pgaudet:
Done. Waiting for build.

PR #2441, opened 2025-02-21
For #2418 do not escape text

mugitty:
Updated to not escape description text since ElementTree,  write method will convert already converted html tags again.  For simple output in aggregate html file, conversion is not required

### DIFF
--- a/scripts/combined_assigned_by.py
+++ b/scripts/combined_assigned_by.py
@@ -15,7 +15,6 @@ import glob
 import copy
 import datetime
 import yamldown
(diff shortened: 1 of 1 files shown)


dustine32 (approved):
@mugitty Approved! Straightforward escape from `html.escape`. If you're able to confirm this change works locally, I'd say go ahead and merge to test in the pipeline.

mugitty:
Thanks @dustine32 , I verified locally with the test.gaf file

PR #2450, opened 2025-03-04
Update retracted-publications.txt

pgaudet:


kltm:
@pgaudet The diff is too large to review here. It would probably be better to get @mugitty to review this one, as she is more familiar with the operative code.

pgaudet:
There is no code - this is just the list of retracted publications, updated yesterday.

pgaudet:
The error message is 
>   note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

I though this was the broader problem... ?

pgaudet:
thanks ! I didn't like the error so much.

PR #2455, opened 2025-03-13
Update goa.yaml

pgaudet:
Removed all isoform, RNA, complexes GPIs and GPADs for human, chicken, dog, pig and cow.

### DIFF
--- a/metadata/datasets/goa.yaml
+++ b/metadata/datasets/goa.yaml
@@ -195,7 +195,7 @@ datasets:
    taxa:
     - NCBITaxon:9606
 
-  -
+ -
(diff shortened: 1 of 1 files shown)


kltm:
@pgaudet Your changes are not valid YAML:
```
ERROR: metadata/datasets/goa.yaml:92:3 [/] document end expected (maybe invalid tab char found).
```

I wanted to do some spot checks as well, but it looks like EBI's HTTPS services are down for the moment and I cannot make any connections. I'll have to check again later.

kltm:
Okay, eyeballing it, I think it's probably good. Let's proceed and find out.

PR #2458, opened 2025-03-20
Update users.yaml

suzialeksander:
https://github.com/geneontology/helpdesk/issues/524

Not sure if these are needed


suzialeksander:
not needed

PR #2462, opened 2025-03-25
Update groups.yaml

suzialeksander:
https://github.com/geneontology/helpdesk/issues/525

PR #2463, opened 2025-03-25
Update users.yaml

suzialeksander:
Depends on info *to be* added in https://github.com/geneontology/go-site/pull/2462 for https://github.com/geneontology/helpdesk/issues/525


suzialeksander:
@kltm does this look OK now?

PR #2465, opened 2025-04-15
Add paint_japonicusdb dataset to paint.yaml

dustine32:
For geneontology/pipeline#413.

This should be merged only after the URL for `gene_association.paint_japonicusdb.gaf.gz` is in-place at `ftp://ftp.pantherdb.org/downloads/paint/presubmission/gene_association.paint_japonicusdb.gaf.gz`.

@kltm @pgaudet Let me know when I can start hosting this new file. Specifically, do you want me to wait until we have a finished `snapshot` run on the current IBA data?

### DIFF
--- a/metadata/datasets/paint.yaml
+++ b/metadata/datasets/paint.yaml
@@ -146,6 +146,23 @@ datasets:
     - NCBITaxon:4896
    merges_into: pombase
 
+-
+   id: paint_japonicusdb.gaf
+   label: "paint_japonicusdb gaf file"
+   description: "gaf file for paint_japonicusdb from PAINT"
+   type: gaf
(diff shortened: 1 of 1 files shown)


kltm:
@dustine32 We are finishing a `snapshot` right now, but either way hosting this additional file is not a problem and it can be added immediately.
That said, at the same time, we'd need the "paint_other" file to have all of the japonicus data removed. As long as these land at the same time, there should be no problem when the data goes up.

dustine32:
@kltm Thanks! I'll merge this and switch the current IBA release to the one containing the new japonicusdb GAF (with the same IBAs removed from `paint_other.gaf`).

Issue #2285, opened 2024-04-04
Merge in metadata changes necessary for using our new (internal) MGI upstreams


kltm:
Our main metadata currently points to the "old" MGI-provided upstreams. Switch to our new / internal files.

sierra-moxon:
PR: https://github.com/geneontology/go-site/pull/2316

pgaudet:
I suppose this is done ?

PR #2508, opened 2025-06-10
add rgd gpi

kltm:
add rgd gpi for #2285


kltm:
Apologies for the duplication, @tutajm .

PR #2516, opened 2025-06-25
Update xenbase.yaml with current GPI

malcolmfisher103:
Updated source for GPI file and xenbase contact email address.


pgaudet:
@malcolmfisher103 Is this the most up-to-date GPI file ? It's from January 2025

PR #2535, opened 2025-07-23
Update db-xrefs.yaml

pgaudet:
Fixed panther links from http to https


kltm:
@pgaudet @balhoff I don't think we can modify the `rdf_uri_prefix` once it's in the wild, correct?
(That said, I'm not sure where this would currently pop out, besides in the triple stores right now?)

PR #2685, opened 2026-06-09
Update URL for Cell Type Ontology

suzialeksander:


suzialeksander:
last fix for https://github.com/geneontology/go-site/issues/2655

PR #2710, opened 2026-06-20
Add --exclude to directory_indexer.py and s3-uploader.py

kltm:
## What

Adds an optional, repeatable `--exclude DIR` flag to both `scripts/directory_indexer.py` and `scripts/s3-uploader.py`. Named directories are skipped entirely — not descended into, and not listed (indexer) / not uploaded (uploader). Implemented by pruning `os.walk`'s `dirs` list in place. Default is empty, so **existing callers are unaffected**.

## Why

The `pipeline-from-goa` publish tail (`scripts/publish-to-s3.sh`) needs to keep the `internal/` staging directory out of *both* the generated directory index and the S3 upload. Neither script currently has an exclude, so the publish script works around it by temporarily relocating `internal/` out of the tree. A native `--exclude internal` lets it drop that workaround. Driver: geneontology/pipeline-from-goa#22.

## Testing

- Both scripts `py_compile` cleanly.
- `directory_indexer.py … --exclude internal` over a sample tree: `internal/` is absent from the root index and no `internal/index.html` is written (it isn't descended into), while sibling directories index normally.
- Backward-compatible: with no `--exclude`, behavior is unchanged (default empty list).

_— Posted by Claude Code agent on behalf of @kltm._

### DIFF
--- a/scripts/bucket-indexer.py
+++ b/scripts/bucket-indexer.py
@@ -113,7 +113,9 @@ def main():
     bucket = s3.Bucket(args.bucket)
     result = bucket.meta.client.list_objects(Bucket=bucket.name, Delimiter='/')
     dirs = []
-    for pre in result.get('CommonPrefixes'):
+    # CommonPrefixes is absent (not []) when the bucket has no '/'-delimited keys,
+    # e.g. an empty bucket -- guard against iterating None.
+    for pre in (result.get('CommonPrefixes') or []):
         name = pre.get('Prefix').rstrip('//')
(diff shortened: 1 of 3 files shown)


kltm:
Added one small robustness fix in the same family while here: `bucket-indexer.py` now guards against `list_objects(Delimiter='/')` omitting `CommonPrefixes` on an empty bucket (it returns `None`, not `[]`, so the loop would `TypeError`). One-line `or []`.

_— Posted by Claude Code agent on behalf of @kltm._

PR #2713, opened 2026-07-01
Update mod_id_space and common names in goex.yaml fixes #2711

### LINKED ISSUE #2711: goex.yaml: mod_id_space / group value inconsistencies (validates, but values drift)
@pgaudet Apologies, when I started to add the comments/descriptions we discussed this morning, I started wondering if I was sure what I said was actually consistent.

It seems that everything is a little off from where I'd expect, and is also used by @dustine32 .

I think it might be best to go from first principles:
- goex.yaml is the source of truth for both sites; we should drive the initial download and download page generation from the goex.yaml and report issues upstream
- While we are currently using `group`, `mod_id_space` would be a better way to key whether something should have a MOD-file as well (?)

Below is a list of detected anomalies.

pgaudet:


kltm:
https://github.com/geneontology/go-site/issues/2711

kltm:
Thanks @pgaudet — the `mod_id_space` changes check out against the actual `-mod` GAFs: `CANAL/CANDC/CANGA/CANPC` are real `CGD`-namespace files (so `CGD` is right), and `CANO9`'s `-mod` is an empty header-only GAF (so `UniProtKB` is right). Two small things before merge:

1. **Trailing space** on `ECO57` (E. coli O157:H7): `code_goc:` → `code_goc: ` — same empty value, just stray whitespace to strip.

2. **`PICGU` `group: UniProt → CGD`** has a downstream side-effect: the downloads-page generator keys its "MOD file" column on `group != UniProt`, so `PICGU` will now get a `PICGU-mod.gaf.gz` link — but there's no `-mod` file for it (404), joining `CANTT`/`CLAL4`. That's really the generator's assumption to fix (geneontology.github.io), not a blocker here — flagging only because the `group` flip triggers it. If CGD is genuinely the authority for `PICGU`, keep the change and we'll fix the generator side.

_— Posted by Claude Code agent on behalf of @kltm._

kltm:
@dustine32 you actually use mod_id_space, right? Technically, "MGI" is the idspace, but there may be practical implications for you. Similarly for uniprot.

kltm:
I want to regenerate pages and this seems at least "more correct", so I'm moving ahead.
@dustine32 I'm not sure how you use uniprot, but for MGI, I think that "MGI" is technically the namespace, it's just that their internally identifiers (incorrectly) have a colon.

PR #2727, opened 2026-08-03
Add gpi files for various Candida species from CGD

pgaudet:
Add CGD GPI files


kltm:
@pgaudet This is my take on it; if it works for you, you can go ahead and merge.

PR #2747, opened 2026-09-17
Add Standard Annotation Backend authorization info to users.yaml schema

### LINKED ISSUE #2746: Add Standard Annotation Backend authorization info to `users.yaml` schema
This task is to update the schema and other QC checks of `users.yaml` to add authorization information for the new Standard Annotation Backend. These will be held in a new `sab` key under the existing `authorizations` key. The `sab` key will hold an array where each item has:

```
role: required; one of `read`, `edit`, or `admin`
scope: required; one of `self`, `group`, or `global`
group: required if scope is `self` or `group`; disallowed if scope is `global`
```

If `group` is specified, it should have the ID of a group listed in `groups.yaml`.

The structural checks can be implemented in the existing kwalify schema in `users.schema.yaml`. The conditional checking can be added to the existing `sanity-check-users-and-groups.py` script.

pkalita-lbl:
Fixes #2746 

* Expands `users.schema.yaml` to allow a new SAB authorization block
* Adds conditional checks that can't be captured by the schema to `sanity-check-users-and-groups.py`
* Adds SAB authorization for myself (SAB is not deployed anywhere yet, but this is needed now for my own local development)


pkalita-lbl:
@kltm since you're away I'm merging now. Forgiveness, permission, yada yada yada

